PR #16, opened 2023-08-20
Feature/generator interface

my-alaska:
Morgan, Atom Pair and Topological Torsion fingerprints implemented with generator interface variant

Handling Sparse output:
Sparse variants of count and bit fingerprints are simply normal count and bit fingerprints converted to 2dimensional scipy csr array sparse arrays

### DIFF
--- a/base.py
+++ b/base.py
@@ -5,13 +5,13 @@ from typing import List, Union
 
 import numpy as np
 import pandas as pd
-from scipy.sparse import vstack, csr_array
+import scipy.sparse as sparse
+
+# from scipy.sparse import vstack, csr_array
 from sklearn.base import BaseEstimator, TransformerMixin
 
 from rdkit.Chem import MolFromSmiles
 from rdkit.Chem.rdchem import Mol
-
-legal_result_types = {"bit", "sparse", "count", "sparse_count"}
 
 
 class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
@@ -22,7 +22,6 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
         """
         self.n_jobs = effective_n_jobs(n_jobs)
         self.fp_generator_kwargs = {}
-        self.concatenate_function = np.concatenate
 
     def fit(self, X, y=None, **fit_params):
         return self
@@ -48,13 +47,15 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
             results = Parallel(n_jobs=self.n_jobs)(
                 delayed(self._calculate_fingerprint)(X_sub) for X_sub in args
             )
-
-            return self.concatenate_function(results)
+            if isinstance(results[0], sparse.csr_array):
+                return sparse.vstack(results)
+            else:
+                return np.concatenate(results)
 
     @abstractmethod
     def _calculate_fingerprint(
         self, X: Union[np.ndarray]
-    ) -> Union[np.ndarray, csr_array]:
+    ) -> Union[np.ndarray, sparse.csr_array]:
         """
         Helper function to be executed in each sub-process.
 
@@ -78,17 +79,16 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
         return X
 
 
-class FingerprintGeneratorInterface(ABC):
-    def __init__(self, result_vector_type: str = "bit"):
+class FingerprintGeneratorMixin(ABC):
+    def __init__(self, fingerprint_type: str = "bit", sparse=False):
         """
         result_vector_tape has to be one of the following:
         bit, spares, count, sparse_count
         """
-        assert result_vector_type in legal_result_types
-        self.result_vector_type = result_vector_type
+        assert fingerprint_type in ["bit", "count"]
+        self.result_vector_type = fingerprint_type
         self.fp_generator_kwargs = {}
-        if result_vector_type in ["sparse", "sparse_count"]:
-            self.concatenate_function = vstack
+        self.sparse_output = sparse
 
     @abstractmethod
     def _get_generator(self):
@@ -99,18 +99,17 @@ class FingerprintGeneratorInterface(ABC):
         """
         pass
 
-    def _sparse_to_csr(self, sparse_vec):
-        return
-
     def _generate_fingerprints(
         self, X: Union[pd.DataFrame, np.ndarray]
-    ) -> Union[np.ndarray, csr_array]:
+    ) -> Union[np.ndarray, sparse.csr_array]:
         fp_generator = self._get_generator()
+
         if self.result_vector_type == "bit":
-            return np.array([fp_generator.GetFingerprint(x) for x in X])
-        elif self.result_vector_type == "sparse":
-            return csr_array([fp_generator.GetFingerprint(x) for x in X])
-        elif self.result_vector_type == "count":
-            return np.array([fp_generator.GetCountFingerprint(x).ToList() for x in X])
+            X = [fp_generator.GetFingerprintAsNumPy(x) for x in X]
         else:
-            return csr_array([fp_generator.GetCountFingerprint(x).ToList() for x in X])
+            X = [fp_generator.GetCountFingerprintAsNumPy(x) for x in X]
+
+        if self.sparse_output:
+            return sparse.csr_array(X)
+        else:
+            return np.array(X)
--- a/base.py
+++ b/base.py
@@ -23,7 +23,6 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
         self.n_jobs = effective_n_jobs(n_jobs)
         self.sparse = sparse
         self.count = count
-        self.fp_generator_kwargs = {}
 
     def fit(self, X, y=None, **fit_params):
         return self
--- a/base.py
+++ b/base.py
@@ -1,13 +1,12 @@
 from abc import ABC, abstractmethod
 
 from joblib import delayed, effective_n_jobs, Parallel
-from typing import List, Union
+from typing import List, Optional, Union
 
 import numpy as np
 import pandas as pd
-import scipy.sparse as sparse
+import scipy.sparse as spsparse
 
-# from scipy.sparse import vstack, csr_array
 from sklearn.base import BaseEstimator, TransformerMixin
 
 from rdkit.Chem import MolFromSmiles
@@ -16,16 +15,14 @@ from rdkit.Chem.rdchem import Mol
 
 class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
     def __init__(
-        self, n_jobs: int = None, sparse=False, fingerprint_type: str = "bit"
+        self,
+        n_jobs: Optional[int] = None,
+        sparse: bool = False,
+        count: bool = False,
     ):
-        """
-        result_vector_type has to be one of the following:
-        bit, sparse, count, sparse_count
-        """
         self.n_jobs = effective_n_jobs(n_jobs)
-        self.sparse_output = sparse
-        assert fingerprint_type in ["bit", "count"]
-        self.result_vector_type = fingerprint_type
+        self.sparse = sparse
+        self.count = count
         self.fp_generator_kwargs = {}
 
     def fit(self, X, y=None, **fit_params):
@@ -52,15 +49,15 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
             results = Parallel(n_jobs=self.n_jobs)(
                 delayed(self._calculate_fingerprint)(X_sub) for X_sub in args
             )
-            if isinstance(results[0], sparse.csr_array):
-                return sparse.vstack(results)
+            if isinstance(results[0], spsparse.csr_array):
+                return spsparse.vstack(results)
             else:
                 return np.concatenate(results)
 
     @abstractmethod
     def _calculate_fingerprint(
         self, X: Union[np.ndarray]
-    ) -> Union[np.ndarray, sparse.csr_array]:
+    ) -> Union[np.ndarray, spsparse.csr_array]:
         """
         Helper function to be executed in each sub-process.
 
@@ -83,7 +80,6 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
         X = [MolFromSmiles(x) if type(x) == str else x for x in X]
         return X
 
-    @abstractmethod
     def _get_generator(self):
         """
         Function that creates a generator object in each sub-process.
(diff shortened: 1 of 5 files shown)


base.py
```diff
@@ -1,20 +1,28 @@
...
 
 import numpy as np
 import pandas as pd
+from scipy.sparse import vstack, csr_array
 from sklearn.base import BaseEstimator, TransformerMixin
 
 from rdkit.Chem import MolFromSmiles
 from rdkit.Chem.rdchem import Mol
 
+legal_result_types = {"bit", "sparse", "count", "sparse_count"}
```
j-adamczyk, line 14: Use UPPER_CASE for global variables

base.py
```diff
@@ -1,20 +1,28 @@
 from abc import ABC, abstractmethod
 
-import joblib
 from joblib import delayed, effective_n_jobs, Parallel
 from typing import List, Union
 
 import numpy as np
 import pandas as pd
+from scipy.sparse import vstack, csr_array
```
j-adamczyk, line 8: Can't we use `import scipy.sparse`? I worry that `vstack` in particular is the same as in Numpy

base.py
j-adamczyk: Also, maybe use `Enum`?

base.py
```diff
@@ -69,3 +76,41 @@ def _validate_input(self, X: List):
 
         X = [MolFromSmiles(x) if type(x) == str else x for x in X]
         return X
+
+
+class FingerprintGeneratorInterface(ABC):
```
j-adamczyk, line 81: Maybe `Mixin` instead of `Interface`, since this is, well, a mixin. And also Scikit-learn uses that convention

base.py
```diff
@@ -69,3 +76,41 @@ def _validate_input(self, X: List):
...
+    def __init__(self, result_vector_type: str = "bit"):
+        """
+        result_vector_tape has to be one of the following:
+        bit, spares, count, sparse_count
+        """
+        assert result_vector_type in legal_result_types
+        self.result_vector_type = result_vector_type
+        self.fp_generator_kwargs = {}
+        if result_vector_type in ["sparse", "sparse_count"]:
+            self.concatenate_function = vstack
```
j-adamczyk, line 91: This is not a good idea. Note that we currently have:
1. Class inherits from `FingerprintTransformer` and `FingerprintGeneratorInterface`.
2. At the beginning of inheriting class, we call parent class constructors in order.
3. `FingerprintTransformer` creates `self.concatenate_function` attribute and assigns default value `np.concatenate`
4. Then, in the line immediately below, `FingerprintGeneratorInterface` constructor is called and it overrides this value

What's more, those two parent classes do not know that other class also uses this attribute! This is a very implicit relationship, and very complicated.

I think we should try to keep to one class, `FingerprintTransformer`. Currently `FingerprintGeneratorInterface` serves 2 purposes (which also breaks Single Responsibility Principle for a mixin, BTW):
1. Provide a way to select concatenation method (`np.concatenate` vs `scipy.sparse.vstack`)
2. Use generator for calculating fingerprints

For concatenating the results, we can just do a simple if/else check at the end, like `if isinstance(results[0], np.ndarray)...`.

To calculate fingerprints with a generator, create a method in `FingerprintTransformer` and just call it in `_generate_fingerprints` in inheriting classes if applicable. And if we want sparse results, we just call `csr_array` on results.

BTW, you don't need a list and NumPy array for generators, you can use `GetFingerprintAsNumPy` and `GetCountFingerprintAsNumPy`.

j-adamczyk (changes requested):

base.py
j-adamczyk: Also, this probably doesn't need to be a global variable

base.py
my-alaska: Modified to import `scipy.sparse as sparse`

base.py
my-alaska: variable removed. the check is made once anyway so it will be performed explicitly as 
`assert result_vector_type in ["bit", "sparse", "count", "sparse_count"]`

base.py
my-alaska: Done

base.py
my-alaska: The concat function has been removed and it's now only using the type check as suggested( *For concatenating the results, we can just do a simple if/else check* )

I'm not sure if I understand correctly how to merge the two classes into one. As for this pull request we haven't implemented the same result vector types in other classes(ones that don't implement generators in rdkit). To prepare for a modification has been introduced so that instead of 4 possible result types passed into FingerprintGeneratorMixin constructor we pass one argument that tells whether we want a bit or count fingerprint and other argument - sparse - bool value that determines if the result is supposed to be a sparse fingerprint or not

Using GetCountFingerprintAsNumpy will help us because it will allow removal of ToList() call but we still have to create a list of numpy fingerprints as both of mentioned methodes can't take whole numpy arrays of molecules. We're only able to pass individual molecules to the function

base.py
```diff
@@ -14,7 +16,12 @@
 
 class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
     def __init__(self, n_jobs: int = None):
+        """
+        result_vector_tape has to be one of the following:
+        bit, spares, count, sparse_count
```
j-adamczyk, line 21: 1. "type", not "tape"
2. "sparse", not "spares"

base.py
```diff
@@ -69,3 +77,39 @@ def _validate_input(self, X: List):
...
+
+        if self.result_vector_type == "bit":
+            X = [fp_generator.GetFingerprintAsNumPy(x) for x in X]
+        else:
+            X = [fp_generator.GetCountFingerprintAsNumPy(x) for x in X]
+
+        if self.sparse_output:
+            return sparse.csr_array(X)
+        else:
+            return np.array(X)
```
j-adamczyk, line 115: I don't get why we can't just have this method in `FingerprintTransformer`. We have everything we need there, I think. I mean, we have information about bit/count, about dense/sparse, and `_get_generator()` was there before anyway.

I get that not all fingerprints have the generator interface, but those ones can just override the `_generate_fingerprints()`, right?

featurizers/fingerprints.py
```diff
@@ -20,216 +21,133 @@
...
-        use_bond_types: bool = True,
-        use_features: bool = False,
+        radius: int = 3,
+        includeChirality: bool = False,
+        useBondTypes: bool = True,
+        onlyNonzeroInvariants: bool = False,
+        includeRingMembership: bool = True,
+        countBounds: AtomPairsParameters = None,
+        fpSize: int = 2048,
+        fingerprint_type: str = "bit",
```
j-adamczyk, line 31: I think we should use Pythonic argument names, not ugly RDKit C++ style ones. After all, this is also why this library exists. And this is relatively small work, isn't it?

featurizers/fingerprints.py
```diff
@@ -20,216 +21,133 @@
...
     def _calculate_fingerprint(
         self, X: Union[pd.DataFrame, np.ndarray]
     ) -> np.ndarray:
+        X = self._validate_input(X)
         from rdkit.Chem.rdMolDescriptors import GetMACCSKeysFingerprint
 
         return np.array([GetMACCSKeysFingerprint(x) for x in X])
 
+    def _get_generator(self):
+        pass
```
j-adamczyk, line 74: If this method is sometimes not used, then it should not be required to be overridden. Currently, this breaks the Interface Segratation Principle rule of SOLID.

tests/fingerprints_tests.py
```diff
@@ -99,73 +105,104 @@ def test_maccs_keys_fingerprint(example_molecules):
...
-    assert np.all(X_atom_pair_hashed == X_seq_hashed)
-    assert np.all(X_atom_pair_as_bit_vect == X_seq_as_bit_vect)
+def test_atom_pair_cound_fingerprint(example_molecules):
+    X = example_molecules
+    X_for_rdkit = [Chem.MolFromSmiles(x) for x in X]
+    rdkit_generator = fpgens.GetAtomPairGenerator()
+    atom_pair = AtomPairFingerprint(fingerprint_type="count", n_jobs=-1)
+    fp_function = rdkit_generator.GetCountFingerprint
+    X_emf = atom_pair.transform(X.copy())
+    X_rdkit = np.array([fp_function(x).ToList() for x in X_for_rdkit])
```
j-adamczyk, line 139: 1. I think that using `GetCountFingerprint` directly would be more readable, i.e.:
```
X_rdkit = np.array([rdkit_generator.GetCountFingerprint(x).ToList() for x in X_for_rdkit])
```
2. Maybe rename `rdkit_generator` to something shorter, e.g. `fp_gen`? In this short context it's readable

j-adamczyk (changes requested):

base.py
```diff
@@ -1,20 +1,32 @@
 from abc import ABC, abstractmethod
 
-import joblib
 from joblib import delayed, effective_n_jobs, Parallel
 from typing import List, Union
 
 import numpy as np
 import pandas as pd
+import scipy.sparse as sparse
```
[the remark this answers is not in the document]
my-alaska: renamed to spsparse
 to avoid overshadowing class constructor argument

base.py
```diff
@@ -1,20 +1,32 @@
...
 from sklearn.base import BaseEstimator, TransformerMixin
 
 from rdkit.Chem import MolFromSmiles
 from rdkit.Chem.rdchem import Mol
 
 
 class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
-    def __init__(self, n_jobs: int = None):
+    def __init__(
+        self, n_jobs: int = None, sparse=False, fingerprint_type: str = "bit"
```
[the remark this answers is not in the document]
my-alaska: Done

featurizers/fingerprints.py
my-alaska: Removed. As well as `@abstractmethod` decorator in base class for the function. Is this convention a good idea? Or should we complete remove the _get_generator function from the base class?

tests/fingerprints_tests.py
my-alaska: done

base.py
```diff
@@ -1,20 +1,32 @@
...
 
-import joblib
 from joblib import delayed, effective_n_jobs, Parallel
 from typing import List, Union
 
 import numpy as np
 import pandas as pd
+import scipy.sparse as sparse
+
+# from scipy.sparse import vstack, csr_array
```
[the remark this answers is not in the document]
my-alaska: removed

featurizers/fingerprints.py
my-alaska: For now changed to Optinal[List]. It's hard to decide whether this is correct as I wasn't able to find many example uses. for the generator. The ones I found typed countBounds as object and also initialized it as none

featurizers/fingerprints.py
```diff
@@ -20,88 +21,51 @@
...
+
+        self.fp_generator_kwargs = {
+            "radius": radius,
+            "includeChirality": includeChirality,
+            "useBondTypes": useBondTypes,
+            "onlyNonzeroInvariants": onlyNonzeroInvariants,
+            "includeRingMembership": includeRingMembership,
+            "countBounds": countBounds,
+            "fpSize": fpSize,
+        }
```
j-adamczyk, line 45: I just realized that this probably won't work. See [Scikit-learn tutorial on custom estimators](https://scikit-learn.org/stable/developers/develop.html#instantiation). In particular, this design will for 100% break cross-validation or Pipeline objects.

I think we're forced to have separate attributes, with Pythonic names. But maybe this won't be that bad - this way `_get_generator()` will be quite explicit.

BTW, we don't have a test for that, I think. We should add generic Scikit-learn tests of estimators, but maybe as a separate PR.

featurizers/fingerprints.py
```diff
@@ -20,88 +21,51 @@
...
-        use_features: bool = False,
-        sparse: bool = False,
-        result_type: str = "default",
+        radius: int = 3,
+        includeChirality: bool = False,
+        useBondTypes: bool = True,
+        onlyNonzeroInvariants: bool = False,
+        includeRingMembership: bool = True,
+        countBounds: Optional[List] = None,
+        fpSize: int = 2048,
```
j-adamczyk, line 30: Change to Pythonic names

base.py
```diff
@@ -1,20 +1,29 @@
...
+    def __init__(
+        self,
+        n_jobs: Optional[int] = None,
+        sparse: bool = False,
+        count: bool = False,
+    ):
         self.n_jobs = effective_n_jobs(n_jobs)
+        self.sparse = sparse
+        self.count = count
+        self.fp_generator_kwargs = {}
```
j-adamczyk, line 26: Remove (reasons outlined below in Morgan fingerprint)

tests/fingerprints_tests.py
```diff
@@ -23,63 +13,77 @@
...
+        "CCN(CCO)CCNc1ccc(C)c2sc3ccccc3c(=O)c12",
+        "Oc1ncnc2c1sc1nc3ccccc3n12",
+        "CC1=CC(=C(c2cc(C)c(O)c(C(=O)O)c2)c2c(Cl)ccc(S(=O)(=O)O)c2Cl)C=C(C(=O)O)C1=O.[NaH]",
     ]
 
 
-def test_morgan_fingerprint(example_molecules):
+def test_morgan_bit_fingerprint(example_molecules):
     X = example_molecules
+    X_for_rdkit = [Chem.MolFromSmiles(x) for x in X]
```
j-adamczyk, line 46: If you make fixtures from this, it will allow you to just pass those to functions as arguments

tests/fingerprints_tests.py
```diff
@@ -23,63 +13,77 @@
...
+    fp_gen = fpgens.GetMorganGenerator()
+    morgan = MorganFingerprint(n_jobs=-1, sparse=False, count=False)
+    X_emf = morgan.transform(X.copy())
```
j-adamczyk, line 49: Do we need to copy X? Why?

j-adamczyk (changes requested):
A few small things left

tests/fingerprints_tests.py
my-alaska: Removed

tests/fingerprints_tests.py
my-alaska: Done

featurizers/fingerprints.py
my-alaska: Done

featurizers/fingerprints.py
my-alaska: Done. I've also added the scikit-lear estimator tests to our backlog

base.py
my-alaska: I've removed the fp_generator_kwargs argument. Why should we also remove sparse and count arguments?

featurizers/fingerprints.py
j-adamczyk: This method in general is ok. Just some classes use it and some don't. I think this is a fair tradeoff

base.py
j-adamczyk: Sorry, my mistake - `sparse` and `count` should stay

j-adamczyk (approved):

PR #17, opened 2023-08-20
Map4

Arch4ngel21:


requirements.txt
```diff
@@ -1,8 +1,12 @@
 black
```
j-adamczyk, line 1: You should add Poetry, sooner than later

requirements.txt
```diff
@@ -1,8 +1,12 @@
...
-ogb
 rdkit
 scikit-learn
 torch-geometric
-tmap
\ No newline at end of file
+e3fp
+mpi4py
+h5py
+standardiser
```
j-adamczyk, line 12: What is this and where are we using this?

requirements.txt
```diff
@@ -1,8 +1,12 @@
...
 PyTDC
-ogb
 rdkit
 scikit-learn
 torch-geometric
-tmap
\ No newline at end of file
+e3fp
+mpi4py
+h5py
```
j-adamczyk, line 11: Why? Are we using this anywhere?

tests/fingerprints_tests.py
```diff
@@ -185,6 +187,23 @@ def test_erg_fingerprint(example_molecules):
     assert np.all(X_erg == X_seq)
 
 
+def test_map4_fingerprint(example_molecules):
+    X = example_molecules
+
+    X_2 = X.copy()
```
j-adamczyk, line 193: Why copy? You create a list of molecules below anyway and assign to `X_2`

tests/fingerprints_tests.py
```diff
@@ -185,6 +187,23 @@ def test_erg_fingerprint(example_molecules):
...
+    # Concurrent
+    map4_fp = MAP4Fingerprint(is_folded=True, n_jobs=-1)
+    X = np.array([Chem.MolFromSmiles(x) for x in X])
+    X_map4 = map4_fp.transform(X.copy())
+
+    # Sequential
+    X_2 = [Chem.MolFromSmiles(x) for x in X_2]
+    X_seq = np.array([GetMAP4Fingerprint(x, is_folded=True) for x in X_2])
+
+    assert np.all(X_map4 == X_seq)
```
j-adamczyk, line 204: Maybe https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html?

tests/fingerprints_tests.py
```diff
@@ -185,6 +187,23 @@ def test_erg_fingerprint(example_molecules):
...
+    X_2 = X.copy()
+
+    # Concurrent
+    map4_fp = MAP4Fingerprint(is_folded=True, n_jobs=-1)
+    X = np.array([Chem.MolFromSmiles(x) for x in X])
+    X_map4 = map4_fp.transform(X.copy())
+
+    # Sequential
+    X_2 = [Chem.MolFromSmiles(x) for x in X_2]
+    X_seq = np.array([GetMAP4Fingerprint(x, is_folded=True) for x in X_2])
```
j-adamczyk, line 202: Why not the same as above, but with `n_jobs=1`?

utils/logger.py
```diff
@@ -0,0 +1,76 @@
...
+
+    with CaptureLogger() as capture:
+        embed_success = AllChem.EmbedMolecule(
+            mol,
+            useExpTorsionAnglePrefs=True,
+            useBasicKnowledge=True,
+            useRandomCoords=True,
+        )
+
+    print(capture["ERROR"])
```
j-adamczyk, line 76: Should this be left here?

j-adamczyk (changes requested):

featurizers/fingerprints.py
```diff
@@ -261,3 +269,40 @@ def _calculate_fingerprint(
...
+
+
+class MAP4Fingerprint(FingerprintTransformer):
+    def __init__(
+        self,
+        dimensions: int = 1024,
+        radius: int = 2,
+        is_counted: bool = False,
+        is_folded: bool = False,
+        return_strings: bool = False,
```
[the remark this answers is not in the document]
Arch4ngel21: This option was present in the original implementation, so I also added it. Actually, it's pointless and will be deleted.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+from rdkit.Chem.rdchem import Mol
+from rdkit.Chem.rdmolops import FindAtomEnvironmentOfRadiusN, GetDistanceMatrix
+
+from featurizers.minhash import Minhash
+
+
+def _find_env(Mol: Mol, idx: int, radius: int) -> str:
+    """
+    Function for getting the environment for given atom. By the environment, we mean the structures,
+    that are adjacent to the given atom.
```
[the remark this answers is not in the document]
Arch4ngel21: It was named after rdkit.Chem.rdmolops.FindAtomEnvironmentOfRadiusN function. I renamed it to "_find_neighborhood()" if it sounds more intuitive.

featurizers/fingerprints.py
```diff
@@ -261,3 +269,40 @@ def _calculate_fingerprint(
...
         return np.array([GetErGFingerprint(x, **fp_args) for x in X])
+
+
+class MAP4Fingerprint(FingerprintTransformer):
+    def __init__(
+        self,
+        dimensions: int = 1024,
+        radius: int = 2,
+        is_counted: bool = False,
+        is_folded: bool = False,
```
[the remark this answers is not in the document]
Arch4ngel21: Then I will separate these 2 fingerprints. In terms of MHFP implementation, I would prefer to keep the version from mhfp library - it is much clearer how MHFPEncoder works.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if idx in atom_map:
+        smiles = MolToSmiles(
+            submol,
+            rootedAtAtom=atom_map[idx],
+            canonical=True,
+            isomericSmiles=False,
+        )
+        return smiles
+    return ""
```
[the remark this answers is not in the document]
Arch4ngel21: This is due to the rdkit.Chem.PathToSubmol function behavior. Unfortunately, I couldn't find a reliable explanation of how this works in documentation/internet, so I tested it myself. Let's take a molecule (with indices in the brackets): C(0) - S(1) - C(2) - O(3) - C(4). Given that, we search for a neighborhood of a radius 2 and we start from atom with index 2, we will get as a result "CSCOC", which is expected. But, if we take a radius 2 and an index 1, the result will be "SCO". It means, that if the path to the most extreme atom is less than the radius, that path won't be included. Also, indices of all atoms at the ends of the molecule will result in an empty string - that is why there is this alternative return "".

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def _get_atom_envs(Mol: Mol, radius: int):
```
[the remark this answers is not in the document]
Arch4ngel21: No, it's a part of the original molecule around the atom of the given index in the SMILES format.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+        atoms_envs = _get_atom_envs(Mol, radius)
+
+        atom_env_pairs = _all_pairs(Mol, atoms_envs, radius, is_counted)
+
+        if is_folded:
+            fp_hash = encoder.hash(set(atom_env_pairs))
+            return encoder.fold(fp_hash, dimensions)
+        elif return_strings:
+            return atom_env_pairs
+        return encoder.from_string_array(atom_env_pairs)
```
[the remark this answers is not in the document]
Arch4ngel21: It will be fixed after separation of MAP4 and MHFP.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+
+    :param Mol:
+    :param radius:
+    :return:
+    """
+    atoms_env = {}
+    for atom in Mol.GetAtoms():
+        idx = atom.GetIdx()
+        for r in range(1, radius + 1):
+            if idx not in atoms_env:
```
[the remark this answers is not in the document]
Arch4ngel21: 1. In the original implementation, there was a `dict` - I don't know if there may exist a situation, in which atom indices are not sequential (there exist some holes in indexing in `Mol`).
2, 3. Changed to `defaultdict()` with a default callable `lambda: []`.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+    :param is_counted:
+    :return:
+    """
+    atom_pairs = []
+    distance_matrix = GetDistanceMatrix(Mol)
+    num_atoms = Mol.GetNumAtoms()
+    shingle_dict = defaultdict(int)
+
+    for idx_1, idx_2 in itertools.combinations(range(num_atoms), 2):
+        dist = str(int(distance_matrix[idx_1][idx_2]))
```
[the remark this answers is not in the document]
Arch4ngel21: Yes, that's the reason. Comment added.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+
+            ordered = sorted([env_a, env_b])
+
+            shingle = "{}|{}|{}".format(ordered[0], dist, ordered[1])
+
+            if is_counted:
+                shingle_dict[shingle] += 1
+                shingle += "|" + str(shingle_dict[shingle])
+
+            atom_pairs.append(shingle.encode("utf-8"))
```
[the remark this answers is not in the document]
Arch4ngel21: Deleted

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+    for idx_1, idx_2 in itertools.combinations(range(num_atoms), 2):
+        dist = str(int(distance_matrix[idx_1][idx_2]))
+
+        for i in range(radius):
+            env_a = atoms_envs[idx_1][i]
+            env_b = atoms_envs[idx_2][i]
+
+            ordered = sorted([env_a, env_b])
+
+            shingle = "{}|{}|{}".format(ordered[0], dist, ordered[1])
```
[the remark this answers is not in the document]
Arch4ngel21: Same molecules can be represented in different ways in SMILES format. Because of that, even for 2 identical molecules, their substructures may be found in a different order, because of a different atom indexing. That part forces order between substructures, which are included in shingles.
Also, changed to f-string.

featurizers/map4.py
```diff
@@ -0,0 +1,128 @@
...
+
+            if is_counted:
+                shingle_dict[shingle] += 1
+                shingle += "|" + str(shingle_dict[shingle])
```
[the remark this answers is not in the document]
Arch4ngel21: Yes, that's incorrect. Changed it, so that if `is_counted`, shingles are stored in `shingle_dict` keys and shingles counts are stored as values. After the nested for-loop, all shingles are counted, so then they're actually added to the `atom_pairs` list with an additional counter at the end.

featurizers/map4.py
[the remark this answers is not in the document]
Arch4ngel21: Same goes for MinHash from datasketch library.

featurizers/minhash.py
```diff
@@ -0,0 +1,49 @@
+"""
+Code taken from https://stackoverflow.com/questions/14533420/can-you-suggest-a-good-minhash-implementation.
+"""
```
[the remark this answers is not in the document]
Arch4ngel21: Added datasketch to requirements and will use this one now instead.

featurizers/map4.py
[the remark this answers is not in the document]
Arch4ngel21: Here's a simple example with a carbon chain C2H5 - CH (CH3) - C5H11:

sample_mol o
sample_mol_2 o

featurizers/fingerprint_helper_functions.py
```diff
@@ -0,0 +1,138 @@
...
+            if not len(env_a) or not len(env_b):
+                continue
+
+            ordered = sorted([env_a, env_b])
+
+            shingle = f"{ordered[0]}|{dist}|{ordered[1]}"
+
+            if is_counted:
+                shingle_dict[shingle] += 1
+                shingle += "|" + str(shingle_dict[shingle])
```
j-adamczyk, line 82: This line doesn't do anything, as far as I can tell. This is the last expression in the inner loop. You first increment counter for a shingle, and then modify the shingle. But then you go into the inner loop again and create completely differnt shingle in line 78.

featurizers/fingerprint_helper_functions.py
```diff
@@ -0,0 +1,138 @@
...
+            else:
+                atom_pairs.append(shingle.encode("utf-8"))
```
j-adamczyk, line 84: Why do we need to encode this as UTF-8? This is a string already, isn't it?

featurizers/fingerprint_helper_functions.py
```diff
@@ -0,0 +1,138 @@
...
+
+    if is_counted:
+        for shingle in shingle_dict.keys():
+            shingle += "|" + str(shingle_dict[shingle])
+            atom_pairs.append(shingle.encode("utf-8"))
```
j-adamczyk, line 89: If I understand correctly, this creates a single huge shingle with counts, right? So something like this?
`sh1a|dist1|sh1b|sh1count sh2a|dist2|sh2b|sh2count sh3a|dist3|sh3b|sh3count ...`

Also, iterate through keys and values (`.items()`). Here you iterate through keys, yet get dictionary values inside the loop.

featurizers/fingerprints.py
```diff
@@ -1,7 +1,11 @@
+import logging
+import sys
 from typing import Union
 
 import numpy as np
 import pandas as pd
+import rdkit.rdBase
```
j-adamczyk, line 7: Is this used anywhere?

featurizers/fingerprints.py
```diff
@@ -1,7 +1,11 @@
+import logging
+import sys
```
j-adamczyk, line 2: Are those used anywhere?

featurizers/fingerprint_helper_functions.py
```diff
@@ -0,0 +1,138 @@
+"""
```
j-adamczyk, line 1: I'm not sure if this should be a one huge file. Maybe we should break fingerprints into separate files? MAP4 and MHFP code is quite large in itself, and `fingerprints.py` file will grow quickly at the current rate.

j-adamczyk (changes requested):

featurizers/fingerprint_helper_functions.py
Arch4ngel21: It was explained a little above in an unresolved comment - Minhash from datasketch library needs this, otherwise it throws an error.

featurizers/fingerprint_helper_functions.py
Arch4ngel21: Yes, that's the idea. Also, changed to `.items()`

featurizers/fingerprint_helper_functions.py
Arch4ngel21: Changed name to `map4_mhfp_helper_functions.py`. The idea from the beginning was, that here would be only helper functions for those 2 fingerprints, but the name was misleading.

base.py
```diff
@@ -38,9 +42,23 @@ def transform(self, X: Union[pd.DataFrame, np.ndarray]):
                 X[i : i + batch_size] for i in range(0, len(X), batch_size)
             )
 
-            results = Parallel(n_jobs=self.n_jobs)(
-                delayed(self._calculate_fingerprint)(X_sub) for X_sub in args
-            )
+            if self.verbose > 0:
+                total_batches = min(self.n_jobs, len(X))
```
j-adamczyk, line 46: Shouldn't this be as follows?
```
total_batches = len(X) // joblib.effective_n_jobs(self.n_jobs)
```

featurizers/fingerprints.py
```diff
@@ -261,3 +265,68 @@ def _calculate_fingerprint(
...
+        self,
+        dimensions: int = 1024,
+        radius: int = 2,
+        is_counted: bool = False,
+        is_folded: bool = False,
+        random_state: int = 0,
+        n_jobs: int = None,
+        verbose: int = 0,
+    ):
+        super().__init__(n_jobs, verbose)
```
j-adamczyk, line 281: I suggest passing by name, so that if we change the parent class, we won't have to change all inheriting classes

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+            rootedAtAtom=atom_map[idx],
+            canonical=True,
+            isomericSmiles=False,
+        )
+        return smiles
+    else:
+        return ""
+
+
+def _get_atom_envs(Mol: Mol, radius: int):
```
j-adamczyk, line 37: Return type is `dict`, right?

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    For each atom get its environment.
+    """
+    atoms_env = defaultdict(lambda: [])
```
j-adamczyk, line 41: Shorter: `defaultdict(list)`

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+    for atom in Mol.GetAtoms():
+        idx = atom.GetIdx()
+        for r in range(1, radius + 1):
+            atoms_env[idx].append(_find_neighborhood(Mol, idx, r))
```
j-adamczyk, line 45: I think this is equivalent to:
```
new_values = [_find_neighborhood(Mol, idx, r) for r in range(1, radius + 1)]
atoms_env[idx].extend(new_values)
```
But `.extend()` should be faster than `.append()` in a loop

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+
+    return atoms_env
+
+
+def _all_pairs(Mol: Mol, atoms_envs: dict, radius: int, is_counted: bool):
```
j-adamczyk, line 50: Return type is `list[str]`, right?

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+    """
```
j-adamczyk, line 51: Good job with this function, it is really complicated to get algorithmic things like this right

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+    #
+    # If we want to count the shingles (is_counted), we increment their value in shingle_dict.
+    # After nested for-loop, all shingles, with their respective counts, will be added to atom_pairs list.
+    for idx_1, idx_2 in itertools.combinations(range(num_atoms), 2):
+        # distance_matrix consists of floats as integers, so they need to be converted to integers first
+        dist = str(int(distance_matrix[idx_1][idx_2]))
+
+        for i in range(radius):
+            env_a = atoms_envs[idx_1][i]
+            env_b = atoms_envs[idx_2][i]
```
j-adamczyk, line 71: Maybe we could extract this outside the inner loop? I.e.:
```
    for idx_1, idx_2 in itertools.combinations(range(num_atoms), 2):
        # distance_matrix consists of floats as integers, so they need to be converted to integers first
        dist = str(int(distance_matrix[idx_1][idx_2]))

        env_a = atoms_envs[idx_1]
        env_b = atoms_envs[idx_2]

        for i in range(radius):
            env_a_radius = env_a[i]
            env_b_radius = env_b[i]
```
This should have relatively low impact, but this way we should avoid repeated dictionary lookups

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+                continue
+
+            ordered = sorted([env_a, env_b])
+
+            shingle = f"{ordered[0]}|{dist}|{ordered[1]}"
+
+            if is_counted:
+                shingle_dict[shingle] += 1
+            else:
+                atom_pairs.append(shingle.encode("utf-8"))
```
j-adamczyk, line 83: Sorry if I asked this before, but why are we encoding a regular string here? Shingly should not contain anything apart from SMILES characters anyway

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+
+    if is_counted:
+        # Shingle in a format:
+        # (neighborhood of atom A of radius i) | (distance between atom A and atom B) | \
+        # (neighborhood of atom B of radius i) | (shingle count)
+        for shingle, shingle_val in shingle_dict.items():
+            shingle += "|" + str(shingle_val)
+            atom_pairs.append(shingle.encode("utf-8"))
```
j-adamczyk, line 91: Convert append to extend:
```
new_atom_pairs = [f"{shingle}|{shingle_count}" for shingle, shingle_count in shingle_dict.items()]
atom_pairs.extend(new_atom_pairs)
```

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -0,0 +1,140 @@
...
+    dimensions: int,
+    radius: int,
+    is_counted: bool,
+    random_state: int,
+):
+    # TODO - There are certain molecules, for which this function will return a error:
+    #   https://github.com/Arch4ngel21/emf/issues/13
+    #   So for now it's handled by try/except
+    try:
+        from datasketch.minhash import MinHash
```
j-adamczyk, line 107: We should inform user that `datasketch` is required. Not necessarily in this PR, but add issue for better handling this

j-adamczyk (changes requested):
A few minor optimization changes left, but looks very nice

base.py
Arch4ngel21: I think not (by the way `effective_n_jobs` is already calculated in `self.n_jobs`) - this would result in obtaining a batch size for data. `total_batches` is only used by `tqdm` to set the number of steps in the progress bar , so it should contain the actual number of processes that we use. Either we can use whole `self.n_jobs` or the length of the data is less than it and we use `len(X)`.

featurizers/map4_mhfp_helper_functions.py
Arch4ngel21: Yes, it was mentioned in 2 comments above - it needs to be encoded, otherwise it raises an error "TypeError: Unicode-objects must be encoded before hashing". It's caused by Minhash from datasketch library.

j-adamczyk: We should note this somewhere that it wants bytes, not strings

base.py
j-adamczyk: This makes sense

j-adamczyk (approved):

PR #18, opened 2023-08-20
E3fp

Arch4ngel21:


featurizers/fingerprints.py
```diff
@@ -261,3 +270,99 @@ def _calculate_fingerprint(
...
+        n_jobs: int = 1,
+        verbose: int = 0,
+    ):
+        if "first" not in confgen_params:
+            confgen_params["first"] = fprint_params.get("first", 1)
+        if "first" not in fprint_params:
+            fprint_params["first"] = confgen_params.get("first", 1)
+
+        # Get necessary default values if not specified
+        from e3fp.conformer.generate import (
```
j-adamczyk, line 293: This can be imported at the top of the file, right?

featurizers/fingerprints.py
```diff
@@ -261,3 +270,99 @@ def _calculate_fingerprint(
...
+        confgen_params["rmsd_cutoff"] = confgen_params.get(
+            "rmsd_cutoff", RMSD_CUTOFF_DEF
+        )
+        confgen_params["max_energy_diff"] = confgen_params.get(
+            "max_energy_diff", MAX_ENERGY_DIFF_DEF
+        )
+        confgen_params["forcefield"] = confgen_params.get(
+            "forcefield", FORCEFIELD_DEF
+        )
+        confgen_params["seed"] = confgen_params.get("seed", SEED_DEF)
```
j-adamczyk, line 317: We have `random_state`, it should be used here

featurizers/fingerprints.py
```diff
@@ -261,3 +270,99 @@ def _calculate_fingerprint(
...
+        )
+        confgen_params["rmsd_cutoff"] = confgen_params.get(
+            "rmsd_cutoff", RMSD_CUTOFF_DEF
+        )
+        confgen_params["max_energy_diff"] = confgen_params.get(
+            "max_energy_diff", MAX_ENERGY_DIFF_DEF
+        )
+        confgen_params["forcefield"] = confgen_params.get(
+            "forcefield", FORCEFIELD_DEF
+        )
```
j-adamczyk, line 316: In general, I think that we should use separate parameters. I know that E3FP library is ugly and uses dicts, but we should do better.

j-adamczyk: I mean separate parameters should be passed, and assigned to attributes. This is also necessary, since Scikit-learn relies on that behavior. See https://scikit-learn.org/stable/developers/develop.html

featurizers/fingerprints.py
```diff
@@ -261,3 +270,99 @@ def _calculate_fingerprint(
...
+            for i in range(len(fps)):
+                fps[i].set_prop("Energy", energies[i])
+
+            if self.is_folded:
+                for i in range(len(fps)):
+                    fps[i] = fps[i].fold(self.fold_bits)
+
+            result.extend(fps)
+
+        result = np.array(result, dtype=object)
```
j-adamczyk, line 366: Why is the result type object?

j-adamczyk (changes requested):

featurizers/fingerprints.py
```diff
@@ -261,3 +275,94 @@ def _calculate_fingerprint(
...
+                },
+            )
+
+            # Set a property for each fingerprint, to be able to obtain energy later
+            for i in range(len(fps)):
+                fps[i].set_prop("Energy", energies[i])
+
+            if self.is_folded:
+                for i in range(len(fps)):
+                    fps[i] = fps[i].fold(self.fold_bits)
```
j-adamczyk, line 362: Maybe we can do this in a single loop, with check inside?
```
for i in range(len(fps)):
    fps[i].set_prop("Energy", energies[i])

    if self.is_folded:
        fps[i] = fps[i].fold(self.fold_bits)
```

tests/fingerprints_tests.py
```diff
@@ -185,6 +197,72 @@ def test_erg_fingerprint(example_molecules):
...
+        ],
+        dtype=object,
+    )
+    X_seq = X_seq.flatten()
+
+    if type(X_seq[0]) is list:
+        new_X_seq = []
+        for x_seq in X_seq:
+            for fp in x_seq:
+                new_X_seq.append(fp)
```
j-adamczyk, line 259: Shorter:
```
new_X_seq = [fp for fp in x_seq for x_seq in X_seq]
```

tests/fingerprints_tests.py
```diff
@@ -185,6 +197,72 @@ def test_erg_fingerprint(example_molecules):
...
+        X_seq = np.array(new_X_seq, dtype=object)
+
+    for i in range(len(X_e3fp)):
+        assert tanimoto(X_e3fp[i], X_seq[i])
```
j-adamczyk, line 263: Maybe be more explicit in this assertion, i.e. `== 1`?

j-adamczyk (changes requested):
A few small optimizations suggested, overall looks very good

featurizers/fingerprints.py
```diff
@@ -261,3 +275,94 @@ def _calculate_fingerprint(
...
+        first: int = 1,
+        num_conf: int = NUM_CONF_DEF,
+        pool_multiplier: float = POOL_MULTIPLIER_DEF,
+        rmsd_cutoff: float = RMSD_CUTOFF_DEF,
+        max_energy_diff: float = MAX_ENERGY_DIFF_DEF,
+        force_field: float = FORCEFIELD_DEF,
+        get_values: bool = True,
+        is_folded: bool = False,
+        fold_bits: int = 1024,
+        standardise: bool = True,
```
j-adamczyk, line 295: I still prefer "standardize" (https://dictionary.cambridge.org/dictionary/english/standardize), since US English is more popular for programming, but ok

j-adamczyk (approved):

PR #26, opened 2023-11-10
Fix map4 and mhfp

my-alaska:
Fix to map4 and mhfp fingerprints, implementation uniformity with other classes and their code style, implementation of sparse variant, tests for sparse and count variants. There's still no implementation or uniformity for e3fp as well as work on actual benchmark as current implementation of e3fp has bugs.

### DIFF
--- a/benchmarking/benchmark.py
+++ b/benchmarking/benchmark.py
@@ -30,7 +30,7 @@ dataset_name = "ogbg-molhiv"
 N_SPLITS = 5
 # N_REPEATS - number of test repetitions - for getting average time score
 N_REPEATS = 5
-N_CORES = [i for i in range(1, cpu_count() + 1)]
+N_CORES = [i for i in range(1, cpu_count() + 1)][:8]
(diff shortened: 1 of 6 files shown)


benchmarking/benchmark.py
```diff
@@ -30,7 +30,7 @@
 N_SPLITS = 5
 # N_REPEATS - number of test repetitions - for getting average time score
 N_REPEATS = 5
-N_CORES = [i for i in range(1, cpu_count() + 1)]
+N_CORES = [i for i in range(1, cpu_count() + 1)][:8]
```
j-adamczyk, line 33: Why? If we have more cores, then good

j-adamczyk (approved):
Just one small comment

benchmarking/benchmark.py
my-alaska: Oh right. It was for the purpose of creating a presentation. I'll fix it

PR #27, opened 2023-11-19
Fix E3FP

my-alaska:
We've discovered some issues in E3FP fingerprint.
The fingerprint couldn't handle more than 1 data element/thread as it returned lists of indices that couldn't be aggregated. Current implementation fixes that issue and returns a sparse arrays that can then be easily aggregated in the thread and outside to stack the results. 

However, this implementation doesn't allow us to return the Energy parameter generated by conformers. I don't know how to fix that without redesigning the result function of the fingerprint that returns either sparse vectors or numpy arrays instead of more complex objects.

The issue of output size being potentially bigger than the input is still not fixed

### DIFF
--- a/featurizers/fingerprints.py
+++ b/featurizers/fingerprints.py
@@ -12,7 +12,7 @@ from e3fp.conformer.generate import (
     RMSD_CUTOFF_DEF,
 )
 from e3fp.conformer.util import mol_to_standardised_mol
-from rdkit.Chem import MolToSmiles, Mol, MolFromSmiles
+from rdkit.Chem import Mol, MolFromSmiles, MolToSmiles
 from rdkit.Chem.PropertyMol import PropertyMol
 
 from base import FingerprintTransformer
@@ -320,16 +320,17 @@ class E3FP(FingerprintTransformer):
         pool_multiplier: float = POOL_MULTIPLIER_DEF,
         rmsd_cutoff: float = RMSD_CUTOFF_DEF,
         max_energy_diff: float = MAX_ENERGY_DIFF_DEF,
-        force_field: float = FORCEFIELD_DEF,
+        forcefield: float = FORCEFIELD_DEF,
         get_values: bool = True,
         is_folded: bool = False,
         fold_bits: int = 1024,
         standardise: bool = True,
-        random_state: int = 0,
+        seed: int = 0,
+        sparse: bool = False,
         n_jobs: int = 1,
         verbose: int = 0,
     ):
-        super().__init__(n_jobs=n_jobs, verbose=verbose, sparse=True)
+        super().__init__(n_jobs=n_jobs, verbose=verbose, sparse=sparse)
         self.bits = bits
         self.radius_multiplier = radius_multiplier
         self.rdkit_invariants = rdkit_invariants
@@ -338,16 +339,16 @@ class E3FP(FingerprintTransformer):
         self.pool_multiplier = pool_multiplier
         self.rmsd_cutoff = rmsd_cutoff
         self.max_energy_diff = max_energy_diff
-        self.force_field = force_field
+        self.forcefield = forcefield
         self.get_values = get_values
         self.is_folded = is_folded
         self.fold_bits = fold_bits
         self.standardise = standardise
-        self.random_state = random_state
+        self.seed = seed
 
     def _calculate_fingerprint(
         self, X: Union[pd.DataFrame, np.ndarray]
-    ) -> np.ndarray:
+    ) -> Union[np.ndarray, spsparse.csr_array]:
         from e3fp.conformer.generator import ConformerGenerator
         from e3fp.pipeline import fprints_from_mol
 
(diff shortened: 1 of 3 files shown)


featurizers/fingerprints.py
```diff
@@ -318,16 +320,17 @@ def __init__(
         pool_multiplier: float = POOL_MULTIPLIER_DEF,
         rmsd_cutoff: float = RMSD_CUTOFF_DEF,
         max_energy_diff: float = MAX_ENERGY_DIFF_DEF,
-        force_field: float = FORCEFIELD_DEF,
+        forcefield: float = FORCEFIELD_DEF,
```
j-adamczyk, line 323: In fact in one of my earlier reviews I suggested the change to `force_field`

featurizers/fingerprints.py
```diff
@@ -318,16 +320,17 @@ def __init__(
...
         get_values: bool = True,
         is_folded: bool = False,
         fold_bits: int = 1024,
         standardise: bool = True,
-        random_state: int = 0,
+        seed: int = 0,
```
j-adamczyk, line 328: We should always use `random_state`, like Scikit-learn

featurizers/fingerprints.py
```diff
@@ -336,46 +339,48 @@ def __init__(
...
-                smiles=MolToSmiles(x),
-                name=MolToSmiles(x),
-                standardise=self.standardise,
-            )
+            if isinstance(x, Mol):
+                smiles = MolToSmiles(x)
+                mol = x
+            else:
+                smiles = x
+                mol = MolFromSmiles(x)
```
j-adamczyk, line 374: According to typing, X is `Union[pd.DataFrame, np.ndarray]`. This assumes that it is a list of either strings or Mol objects. I think that we assume the former in all cases, i.e. we work more like `CountVectorizer` than `PolynomialFeatures` transformer. Right? So the typing should be changed here.

featurizers/fingerprints.py
```diff
@@ -336,46 +339,48 @@ def __init__(
...
+
+            mol.SetProp("_Name", smiles)
+            if self.standardise:
+                mol = mol_to_standardised_mol(mol)
```
j-adamczyk, line 378: Standardization of molecules is a very typical preprocessing step and while we should probably make a utility function (as a transformer class) of it, it shouldn't be managed by this class. Standardized molecules can be used for any fingerprints, after all.

featurizers/fingerprints.py
```diff
@@ -386,15 +391,15 @@ def _calculate_fingerprint(
...
-                fps[i].set_prop("Energy", energies[i])
-
-                if self.is_folded:
-                    fps[i] = fps[i].fold(self.fold_bits)
+            energies = values[2]
+            fp = fps[np.argmin(energies)]
 
-            result.extend(fps)
+            if self.is_folded:
+                fp = fp.fold(self.fold_bits)
```
j-adamczyk, line 398: I wonder if we should allow non-folded version at all. After all, we assume Scikit-learn interface, which does require Numpy or Scipy matrices.

featurizers/fingerprints.py
```diff
@@ -386,15 +391,15 @@ def _calculate_fingerprint(
...
             )
 
-            # Set a property for each fingerprint, to be able to obtain energy later
-            for i in range(len(fps)):
-                fps[i].set_prop("Energy", energies[i])
-
-                if self.is_folded:
-                    fps[i] = fps[i].fold(self.fold_bits)
+            energies = values[2]
+            fp = fps[np.argmin(energies)]
```
j-adamczyk, line 395: Make this an argument to the constructor, e.g. `aggregation_type`, with default value `min_energy` or something like that. This should be expanded in the future, after all.

j-adamczyk (changes requested):
A few small suggestions left, looks very good

featurizers/fingerprints.py
my-alaska: Done

featurizers/fingerprints.py
my-alaska: done. also moved to base class

featurizers/fingerprints.py
my-alaska: Current implementation assumes both smiles strings and mol objects. We can process smiles and turn them into molecules with use of _validate_input() from the base class. This allows us to not only compute the fingerprints concurrently but also the transformation from smiles to fingerprints, as it turned out to be slow and could potentially become a bottleneck for larger datasets. I think we decided to do that after finding out that initial benchmark scores didn't result in a great time improvement

my-alaska: We could create a new SmilesToMol transformer but the time effort for that would most likely be greater than we can afford before our deadlines. I'll be happy to implement it as future work

featurizers/fingerprints.py
my-alaska: Removed. 

We could implement a standarisation helper function in the base class that would work similarily to _validate_input and that could also allow to standarise molecules concurrently. Also TODO as future work?

featurizers/fingerprints.py
my-alaska: `fprints_from_mol()` takes number of bits as an argument and limits the upper bound of the output tensor to that number. Folding is then performed on vectors of already limited and uniform size

featurizers/fingerprints.py
my-alaska: Added an "if" statement for that. The "else" choses the first fingerprint in the list as a placeholder

j-adamczyk (approved):

PR #40, opened 2023-11-28
Benchmark

my-alaska:
MAP4 fixed, returns uniform vectors of size 128(can be changed). This allows for proper concatenation of numpy or sparse vectors.
Added count input parameter so that the interfeaces are uniform with the base class.
Many changes to details like names, initialized parameters etc.

### DIFF
--- a/featurizers/fingerprints.py
+++ b/featurizers/fingerprints.py
@@ -190,13 +190,14 @@ class MACCSKeysFingerprint(FingerprintTransformer):
         n_jobs: int = 1,
         verbose: int = 0,
         random_state: int = 0,
-        count: Optional[bool] = None,  # unused
+        count: bool = False,  # unused
     ):
         super().__init__(
             n_jobs=n_jobs,
             sparse=sparse,
             verbose=verbose,
             random_state=random_state,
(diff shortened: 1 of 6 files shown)


featurizers/fingerprints.py
```diff
@@ -348,8 +351,8 @@ def _calculate_fingerprint(
 class E3FP(FingerprintTransformer):
     def __init__(
         self,
-        bits: int,
-        radius_multiplier: float,
+        bits: int = 4090,
```
j-adamczyk, line 354: Why such a weird number? In E3FP paper they use 1024 as default number of bits

featurizers/map4_mhfp_helper_functions.py
```diff
@@ -103,7 +103,7 @@ def _all_pairs(
 
 def get_map4_fingerprint(
     Mol: Mol,
-    dimensions: int = 1024,
+    dimensions: int = 128,
```
j-adamczyk, line 106: Why? 1024 is the default value in MAP4 paper

j-adamczyk (approved):
2 comments left

PR #63, opened 2023-12-13
Create python-publish.yml

przemekk6973:


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,40 @@
...
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
```
j-adamczyk, line 15: Remove emojis and shorten to "Buil and publish to PyPI" if possible

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,40 @@
...
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries
+
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+
+on: push
```
j-adamczyk, line 11: We don't want to publish on every master push, only when we tag a release. Maybe also only from a separate release branch. How is it handled here?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  build-n-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-18.04
```
j-adamczyk, line 16: This is an old system. Can we run on 22, or at least 20?

.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,40 @@
...
+        poetry
+        --user
+    - name: Configure poetry
+      run: poetry config pypi-token.pypi ${{ secrets.PYPI_API_TOKEN }}
+    - name: Build and publish to TestPyPI
+      run: >-
+        poetry build
+        && poetry publish --repository testpypi
+    - name: Build and publish to PyPI
+      if: startsWith(github.ref, 'refs/tags')
```
j-adamczyk, line 36: This ensures that we only build and publish tagged releases?

j-adamczyk (changes requested):

przemekk6973:
The publish on release action works, but testpypi doesn't allow to publish the same version of the library twice. Thats why the latest run failed. However, there is a succesful run in the actions history and when the newer version of library is published, everything will be fine. As far as the test scripts are concerned, after last week's changes, the e3fp tests do not work. We agreed with Piotrek to remove them for now, as the fingerprint needs fix in the future anyway. Once the test scripts are done, this PR is ready to merge.
image

pyproject.toml
```diff
@@ -23,6 +23,7 @@ datasketch = "*"
 e3fp = "*"
 joblib = "*"
 mhfp = "*"
+ogb = "*"
```
j-adamczyk, line 26: This should not be a requirement. We do not use OGB in the library itself

j-adamczyk (changes requested):

PR #72, opened 2023-12-24
E3FP fix and RDK fingerprint

Arch4ngel21:


benchmarking/benchmark.py
```diff
@@ -225,15 +226,19 @@ def get_times_e3fp(X: pd.DataFrame):
...
+                    # getting a molecule and the fingerprint
+                    mol, values = conf_gen.generate_conformers(mol)
+                    fps = fprints_from_mol(mol, fprint_params=fprint_params)
 
-                # chose the fingerprint with the lowest energy
-                energies = values[2]
-                fp = fps[np.argmin(energies)].fold(1024)
+                    # chose the fingerprint with the lowest energy
+                    energies = values[2]
+                    fp = fps[np.argmin(energies)].fold(1024)
```
j-adamczyk, line 236: Do not use constant 1024, this is a constructor parameter

benchmarking/benchmark.py
```diff
@@ -225,15 +226,19 @@ def get_times_e3fp(X: pd.DataFrame):
...
+
+                    X_seq.append(fp.to_vector())
+                except RuntimeError:
+                    X_seq.append(np.full(shape=1024, fill_value=-1))
```
j-adamczyk, line 240: Do not use constant 1024, this is a constructor parameter

j-adamczyk: Also, maybe return all zeros instead of all -1?

skfp/fingerprints/rdk.py
```diff
@@ -0,0 +1,61 @@
+from typing import List, Optional, Union
+
+import numpy as np
+import pandas as pd
+import scipy.sparse as spsparse
+
+from skfp.fingerprints.base import FingerprintTransformer
+
+
+class RDKFingerprint(FingerprintTransformer):
```
j-adamczyk, line 10: I would all this fully `RDKitFingerprint`, since this is the official name

j-adamczyk (changes requested):

benchmarking/benchmark.py
Arch4ngel21: But isn't this a better way to handle such cases? Vector of -1's is easy to find and, e.g., to be excluded later by the user, while full of zeros is not very apparent and may lead to hidden false information in the data.

benchmarking/benchmark.py
Arch4ngel21: IMO it's better to leave like this. Either 1. we'd need to define a hyperparameter for folding bits - that means we should pass it as an argument to all fingerprint constructors (that requires it) and that is messing with the current helper function for calculating time for our library, or 2. we'd need to use the parameter's value from E3FP instance, which is not directly accessible from the function above.

benchmarking/benchmark.py
Arch4ngel21: MAP4, MHFP and E3FP have handling like this (with vector of -1's)

j-adamczyk: Ok

benchmarking/benchmark.py
j-adamczyk: The target dimensionality is the number of bits after folding. I mean, if the default value for E3FP dimensionality is 1024, then it should be 1024 after folding. But for now let's leave this as-is, we can change it later if needed.

j-adamczyk (approved):

PR #77, opened 2024-01-25
e-state fingerprint

my-alaska:
EState fingerprint functionality.

Please note that the fingerprint returns 2 arrays. This is how it was implemented in RDKit.
Each fingerprint is returned as a Python tuple of 2 NumPy arrays of the same size. Maybe could concatenate them instead, or specify return type as "count", "sum" or "both"?

### DIFF
--- a/skfp/fingerprints/e_state.py
+++ b/skfp/fingerprints/e_state.py
@@ -8,11 +8,6 @@ from skfp.fingerprints.base import FingerprintTransformer
 
 
 class EStateFingerprint(FingerprintTransformer):
-    """this fingerprint returns 2 arrays:
-    The first (of ints) contains the number of times each possible atom type is hit
-     The second (of floats) contains the sum of the EState indices for atoms of
-    """
-
     def __init__(
         self,
         variant: str = "sum",
@@ -22,9 +17,7 @@ class EStateFingerprint(FingerprintTransformer):
         random_state: int = 0,
         count: bool = False,
     ):
-        """'variant' argument determines the output type. It can be one of:
-        'sum', 'count' or 'binary'
-        """
+        """'variant' argument determines the output type. It can be one 'sum', or 'binary'"""
         super().__init__(
             n_jobs=n_jobs,
             sparse=sparse,
@@ -32,6 +25,12 @@ class EStateFingerprint(FingerprintTransformer):
             random_state=random_state,
             count=count,
         )
+        if variant not in ["sum", "binary"]:
+            raise ValueError("variant must be 'sum' or 'binary'")
+        if self.count and variant == "sum":
+            raise ValueError(
+                "count fingerprint only available with 'binary' variant"
+            )
         self.variant = variant
 
     def _calculate_fingerprint(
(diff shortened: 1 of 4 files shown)


skfp/fingerprints/e_state.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    def __init__(
+        self,
+        variant: str = "sum",
+        sparse: bool = False,
+        n_jobs: int = None,
+        verbose: int = 0,
+        random_state: int = 0,
+        count: bool = False,
```
j-adamczyk, line 23: How does the combination of `variant` and `count` work? I think we should raise an error for `variant="sum"` with `count=True`, since this combination doesn't make sense.

skfp/fingerprints/e_state.py
```diff
@@ -0,0 +1,54 @@
...
+import scipy.sparse as spsparse
+
+from skfp.fingerprints.base import FingerprintTransformer
+
+
+class EStateFingerprint(FingerprintTransformer):
+    """this fingerprint returns 2 arrays:
+    The first (of ints) contains the number of times each possible atom type is hit
+     The second (of floats) contains the sum of the EState indices for atoms of
+    """
```
j-adamczyk, line 14: This is no longer true

skfp/fingerprints/e_state.py
```diff
@@ -0,0 +1,54 @@
...
+        self,
+        variant: str = "sum",
+        sparse: bool = False,
+        n_jobs: int = None,
+        verbose: int = 0,
+        random_state: int = 0,
+        count: bool = False,
+    ):
+        """'variant' argument determines the output type. It can be one of:
+        'sum', 'count' or 'binary'
```
j-adamczyk, line 26: If we keep the `count` argument, then `variant` can be only `sum` or `count`

skfp/fingerprints/e_state.py
```diff
@@ -0,0 +1,54 @@
...
+        """
+        super().__init__(
+            n_jobs=n_jobs,
+            sparse=sparse,
+            verbose=verbose,
+            random_state=random_state,
+            count=count,
+        )
+        self.variant = variant
```
j-adamczyk, line 35: We should do the check here for `variant` and `count`, as explained above

skfp/fingerprints/e_state.py
my-alaska: That's the problematic part. We have the `count` member of the parent class and `count` option for `variant` and that introduces a redundancy problem. I think we could have 2 possible solutions to that. One would be removing the original `count` from the constructor and initializing it as false in the parent class constructor call. The other option would be to ignore the `variant` when `count` is true and remove it as a legal variant, leaving only `"sum"` and `"binary"`

my-alaska: Nevermind. I read comments below and the `count` set to `true` will be legal with `"binary"` variant

j-adamczyk (approved):

PR #78, opened 2024-02-11
MHFP rdkit implementation

my-alaska:
Faster MHFP implementation. The previous implementation is left as MHFPlegacy. It appears that the comments and documentation for EncodeMolsBulk function are incorrect as it states that `kekulize` parameter is initialized as false as in reality it's true.


j-adamczyk (approved):
I think we can remove old MHFP, we want to rely on RDKit for this. Aside from this, LGTM

PR #79, opened 2024-02-20
mol to smiles and mol from smiles

my-alaska:
Added MolFromSmiles and MolToSmiles transformers

### DIFF
--- a/skfp/fingerprints/base.py
+++ b/skfp/fingerprints/base.py
@@ -39,6 +39,10 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
         self.verbose = verbose
         self.random_state = random_state
 
+        # value that helps to tell apart fingerprint and utility transformers
+        # set as True for util transformers
+        self.fingerprint_transformer = True
+
     def fit(self, X, y=None, **fit_params):
         return self
 
@@ -78,7 +82,9 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
                     for X_sub in args
                 )
 
-            if self.sparse:
+            if not self.fingerprint_transformer:
+                return pd.concat(results)
+            elif self.sparse:
                 return spsparse.vstack(results)
             else:
                 return np.vstack(results)
--- a/skfp/fingerprints/base.py
+++ b/skfp/fingerprints/base.py
@@ -39,10 +39,6 @@ class FingerprintTransformer(ABC, TransformerMixin, BaseEstimator):
         self.verbose = verbose
         self.random_state = random_state
 
-        # value that helps to tell apart fingerprint and utility transformers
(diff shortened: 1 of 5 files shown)


skfp/fingerprints/base.py
```diff
@@ -39,6 +39,10 @@ def __init__(
         self.verbose = verbose
         self.random_state = random_state
 
+        # value that helps to tell apart fingerprint and utility transformers
+        # set as True for util transformers
+        self.fingerprint_transformer = True
```
j-adamczyk, line 44: We should just create separate parent classes. Or, better, just don't use parent classes at all for those utility transformers.

skfp/other/util_transformers.py
```diff
@@ -0,0 +1,91 @@
...
+
+from skfp.fingerprints.base import FingerprintTransformer
+
+
+class MolFromSmilesTransformer(FingerprintTransformer):
+    def __init__(
+        self,
+        sanitize: bool = True,
+        replacements: dict = {},
+        random_state: int = 0,
```
j-adamczyk, line 16: There is no randomness here

skfp/other/util_transformers.py
```diff
@@ -0,0 +1,91 @@
...
+            count=False,
+            verbose=verbose,
+            random_state=random_state,
+        )
+        self.sanitize = sanitize
+        self.replacements = replacements
+        self.fingerprint_transformer = False
+
+    def _calculate_fingerprint(
+        self, X: Union[pd.DataFrame, list[str]]
```
j-adamczyk, line 33: This won't work for `pd.DataFrame`, nor for `pd.Series`, as it is. I would leave just `list[str]`

skfp/other/util_transformers.py
```diff
@@ -0,0 +1,91 @@
...
+    ) -> pd.Series:
```
j-adamczyk, line 34: Return list of molecules. Nothing in Scikit-learn returns pd.Series

skfp/other/util_transformers.py
```diff
@@ -0,0 +1,91 @@
...
+        self.kekule_smiles = kekule_smiles
+        self.rooted_at_atom = rooted_at_atom
+        self.canonical = canonical
+        self.all_bonds_explicit = all_bonds_explicit
+        self.all_hs_explicit = all_hs_explicit
+        self.do_random = do_random
+
+        self.fingerprint_transformer = False
+
+    def _calculate_fingerprint(self, X: list[Mol]) -> pd.Series:
```
j-adamczyk, line 76: Return regular list instead of pd.Series

j-adamczyk (changes requested):

j-adamczyk (approved):

PR #82, opened 2024-03-09
CICD and dev experience improvements

mjste:
This feature brings up several issues:

* **Publish to PyPI** — Change to constantly failing publish action. Now it should only run this action only on closing pull request from branches named like `release/**`
* **Poetry only** — Unifies dependency management. No mixing of poetry and pip, all used dependencies are present in `pyproject.toml`
* **Lighter (faster) tests** are added. On common push, we are checking only part of the dataset instead of the whole. This way, action takes only a few minutes instead of hours. Shorter, but still valuable, feedback loop.
* Tests are triggered only on code change. `Changes in README.md`, `CONTRIBUTING.md` should not run tests.
* * Changes poetry installation way in actions: previous method is discouraged according to poetry docs. Current one uses separate poetry installed in venv.
*  **CONTRIBUTING.md** — added info about pipx and pre-commit hooks


CONTRIBUTING.md
```diff
@@ -21,30 +21,35 @@ We encourage you to submit pull requests (PRs) to improve our project. To do so,
    ```bash
    git clone https://github.com/Arch4ngel21/scikit-fingerprints.git
 
-3. If you don't have it, install poetry:
+3. If you don't have it, install poetry (preferably with pipx):
    ```bash
-   pip install poetry
+   pipx install poetry
```
j-adamczyk, line 26: I wouldn't explicitly mention this, since this is not directly relevant to the library. Also, what if Poetry changes something here? I think avoiding such explicit dependencies would be a good idea

pyproject.toml
```diff
@@ -31,9 +31,24 @@ scikit-learn = "*"
...
+[tool.poetry.group.dev]
+optional = true
+
 [tool.poetry.group.dev.dependencies]
 flake8 = "*"
 pytest = "*"
+pre-commit = "*"
+black = "*"
+isort = "*"
+mypy = "*"
```
j-adamczyk, line 43: Sort alphabetically. Maybe we should also use https://github.com/andrei-shabanski/poetry-plugin-sort for development?

CONTRIBUTING.md
mjste: I thought of mentioning it, because it is explicitly advised against using bare pip to install poetry in official poetry docs [(reference (red ban sign))](https://python-poetry.org/docs/#installation).
Mentioning pipx may not be the best idea, but I still think that it is better than suggesting doing explicitly discouraged actions. Alternatively, we could just mention that poetry is required, with no suggestion how to install it

pyproject.toml
mjste: Done.

CONTRIBUTING.md
j-adamczyk: Yes, I think we should not mention anything besides that Poetry is required. We shouldn't care how developer installs it.

j-adamczyk (approved):

PR #83, opened 2024-03-16
Major refactoring and cleanup

j-adamczyk:


pyproject.toml
```diff
@@ -23,9 +23,9 @@ python = "^3.9"
 datasketch = "*"
 e3fp = "*"
 joblib = "*"
-mhfp = "*"
 numpy = "^1.20.0"
 pandas = "*"
+pytest = "^8.1.1"
```
mjste, line 28: Why do we have pytest dependency added twice? (Second in test group dependencies)

j-adamczyk: Good catch, I removed it

mjste (changes requested):
Tests are failing after last commit

j-adamczyk:
@mjste fixed

PR #85, opened 2024-03-16
Add multiple descriptor-based fingerprints from RDKit

j-adamczyk:
Adds:
- autocorrelation
- MORSE
- RDF
- WHIM
- 3D pharmacophore

Also contains 3D conformer generation required for those.

I had problems with BCUT and GETAWAY, described in issues on RDKit Github:
- https://github.com/rdkit/rdkit/issues/7265
- https://github.com/rdkit/rdkit/issues/7264


skfp/preprocessing.py
```diff
@@ -64,3 +65,67 @@ def transform(self, X):
...
+                    pass
+
+            if conf_id == -1:
+                try:
+                    # turn off conditions
+                    embed_params.enforceChirality = False
+                    embed_params.ignoreSmoothingFailures = True
+                    conf_id = EmbedMolecule(mol, embed_params)
+                except ValueError:
+                    pass
```
my-alaska, line 115: Wouldn't it make sense to replace these 2 blocks with one if statement and try/except? I don't think it would impact readability

j-adamczyk: I tried that first, but it results in more indents and was less readable in my opinion. I further simplified this file now

my-alaska (approved):

PR #91, opened 2024-03-19
Add physiochemical (binding) properties fingerprint

j-adamczyk:


skfp/fingerprints/base.py
```diff
@@ -95,3 +95,16 @@ def _validate_input(self, X: List, smiles_only: bool = False) -> List[Mol]:
 
         X = [MolFromSmiles(x) if isinstance(x, str) else x for x in X]
         return X
+
+    def _hash_fingerprint_bits(self, X: List) -> Union[np.ndarray, csr_array]:
+        shape = (len(X), self.fp_size)
```
my-alaska, line 100: Maybe it would be worth to create a self.fp_size field in the base class and initialize as 0? Just for the sake of consistency as so far, the base class only referenced its own member variables.

tests/fingerprints/physiochemical_properties.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+from scipy.sparse import csr_array
+
+from skfp.fingerprints import PhysiochemicalPropertiesFingerprint
+
+
+def test_physiochemical_properties_bp_bit_fingerprint(smiles_list):
```
my-alaska, line 7: Didn't we settle for removing the "bit" part of test names?

tests/fingerprints/physiochemical_properties.py
```diff
@@ -0,0 +1,100 @@
...
+    pp_fp = PhysiochemicalPropertiesFingerprint(
+        variant="BP", sparse=False, count=False, n_jobs=-1
+    )
+    X_skfp = pp_fp.transform(smiles_list)
+
+    assert isinstance(X_skfp, np.ndarray)
```
my-alaska, line 13: Many security tools won't allow for the use of `assert` and instead they'll insist on using 
`if not condition: raise AssertionError ("[rror message")`

my-alaska (changes requested):

skfp/fingerprints/base.py
j-adamczyk: This implicitly assumes that inheriting class has `fp_size` attribute, that's true. But I don't think this is avoidable, since we want all fingerprints to inherit from FingerprintTransformer, and all those that need hashing will have `fp_size`.

I will add an explicit check for those attributes, to avoid problems.

tests/fingerprints/physiochemical_properties.py
j-adamczyk: Well, here we have both bit and count variants, so I left this, to be explicit. Only those fingerprints that have only a single variant don't have this in test names.

tests/fingerprints/physiochemical_properties.py
j-adamczyk: This does not apply to tests

my-alaska (approved):

PR #107, opened 2024-03-28
Reduce E3FP logs

j-adamczyk:


skfp/fingerprints/e3fp_fp.py
```diff
@@ -91,15 +93,23 @@ def _calculate_single_mol_fingerprint(
...
-            mol, values = conf_gen.generate_conformers(mol)
-            fps = fprints_from_mol(
-                mol,
-                fprint_params={
-                    "bits": self.n_bits_before_hash,
-                    "radius_multiplier": self.radius_multiplier,
-                    "rdkit_invariants": self.rdkit_invariants,
-                },
-            )
+            try:
```
mjste, line 96: Please add a switch instead of disabling it at all

mjste (changes requested):

skfp/fingerprints/e3fp_fp.py
my-alaska: +1

mjste (approved):

my-alaska (approved):

PR #110, opened 2024-03-29
Add LayeredFingeprint

j-adamczyk:


mjste (approved):

skfp/fingerprints/layered.py
```diff
@@ -0,0 +1,50 @@
...
+        branched_paths: bool = True,
+        sparse: bool = False,
+        n_jobs: Optional[int] = None,
+        verbose: int = 0,
+    ):
+        super().__init__(
+            sparse=sparse,
+            n_jobs=n_jobs,
+            verbose=verbose,
+        )
```
my-alaska, line 28: It might be wise to already add the n_features parameter here from the [set_output pull request](https://github.com/scikit-fingerprints/scikit-fingerprints/pull/113) to avoid changes later

my-alaska (changes requested):

skfp/fingerprints/layered.py
j-adamczyk: This parameter is not available yet. Let's merge that PR first and then this one

PR #113, opened 2024-03-31
Honor set_output option from scikit-learn

j-adamczyk:


skfp/fingerprints/mordred_fp.py
```diff
@@ -19,7 +19,9 @@ def __init__(
         n_jobs: Optional[int] = None,
         verbose: int = 0,
     ):
+        n_feaatures_out = 1826 if use_3D else 1613
```
my-alaska, line 22: Older version of the Mordred fingerprint had different numbers of output features. Even though we don't expect the user to use them, we can notify them about it in a code comment here. ([info](https://github.com/mordred-descriptor/mordred/blob/develop/README.rst))

mjste (approved):

skfp/fingerprints/mordred_fp.py
j-adamczyk: I added explicit output shape tests to all fingerprints to ensure that we don't miss such cases

PR #115, opened 2024-04-01
Expand GitHub Actions tests

j-adamczyk:


my-alaska (approved):

.github/workflows/python-test.yml
```diff
@@ -1,40 +1,49 @@
...
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
 
-    - name: Set up Python
-      uses: actions/setup-python@v4
-      with:
-        python-version: '3.9'
+      - name: Install Poetry
+        uses: abatilo/actions-poetry@v2
```
mjste, line 26: Version 3 is available. Can we use it?

mjste, line 26: We can probably add sort plugin here as a parameter to this action instead of our own step

j-adamczyk: Sure

j-adamczyk: Good idea

mjste (approved):

mjste (approved):

PR #117, opened 2024-04-01
Add Atom Pair docs and fixes

j-adamczyk:


docs/make.bat
mjste, line 1: maybe we should leave CRLF, as .bat files are meant for windows

docs/_templates/class.rst
mjste, line 1: move it to _templates/autosummary/class.rst
Without it, the file does not affect anything

docs/_templates/class.rst
```diff
@@ -0,0 +1,27 @@
...
+
+.. autoclass:: {{ objname }}
+   {% block methods %}
+
+   {% if methods %}
+   .. rubric:: {{ _('Methods') }}
+
+   .. autosummary::
+   {% for item in methods %}
+      ~{{ name }}.{{ item }}
```
mjste, line 13: please add 
`
{% if item != '__init__' %} 
  ~{{ name }}.{{ item }}
{% endif %}
`

mjste (changes requested):
at least class.rst must be moved

docs/make.bat
j-adamczyk: done

docs/_templates/class.rst
j-adamczyk: done

mjste (approved):

PR #121, opened 2024-04-02
Add parameter constraints everywhere

j-adamczyk:


tests/fingerprints/topological_torsion.py
mjste, line 1: 1) Should we add count to our interface if it is not implemented in RDKit?
2) If not, we do not need so many tests, because they are effectively duplicates

Is it here to inform us whether RDKIT has implemented it?

j-adamczyk: We have count option everywhere that RDKit also does, and nowhere else. In particular, RDKit has count option for Topological Torsion FP.

mjste: my bad

Issue #80, opened 2024-03-05
Extend Atom Pairs fingerprint with HAC scaling


j-adamczyk:
Heavy Atom Scaling (HAC) rescales feature counts for Atom Pairs fingerprint, making them proportional to the molecule size. This can be a useful hyperparameter.

"Atom pair 2D-fingerprints perceive 3D-molecular shape and pharmacophores for very fast virtual screening of ZINC and GDB-17" M. Awale, J. Reymond
> Similarly to the approaches by Carhart and Schneider, the counts were scaled to the heavy atom count (HAC) to obtain a simple proportionality to molecular size since the number of possible atom pairs increases with the square of HAC. Bit values were expressed in percent and rounded to the integer value.

PR #123, opened 2024-04-02
Add atom pair normalization

BlazejNowicki:
[Linked issue](https://github.com/scikit-fingerprints/scikit-fingerprints/issues/80)


skfp/fingerprints/atom_pair.py
```diff
@@ -63,6 +64,11 @@ class AtomPairFingerprint(FingerprintTransformer):
     count : bool, default=False
         Whether to return binary (bit) features, or their counts.
 
+    normalize: bool, default=False
+        Wheather to scale count fingerprint by the heavy atom count (HAC) to
```
j-adamczyk, line 68: Typo: "Whether"

skfp/fingerprints/atom_pair.py
```diff
@@ -63,6 +64,11 @@ class AtomPairFingerprint(FingerprintTransformer):
...
+        obtain a proportionality to molecular size. Vector values are
```
j-adamczyk, line 69: "to molecule size", I think, since we are talking about a particular molecule, not "molecular" in general

skfp/fingerprints/atom_pair.py
```diff
@@ -63,6 +64,11 @@ class AtomPairFingerprint(FingerprintTransformer):
...
+        expressed in percent and rounded to the nearest integer.
```
j-adamczyk, line 70: Maybe "Values are expressed as percentages, rounded to the nearest integer"?

j-adamczyk: Also, add reference to the paper that mentioned HAC scaling here, I think it's already in references list

skfp/fingerprints/atom_pair.py
```diff
@@ -206,14 +215,31 @@ def _calculate_fingerprint(
             countSimulation=self.count_simulation,
         )
         if self.count:
-            X = [
-                gen.GetCountFingerprintAsNumPy(mol, confId=conf_id)
-                for mol, conf_id in zip(X, conf_ids)
-            ]
+            if self.normalize:
+                X = [
+                    self._scale_by_hac(
```
j-adamczyk, line 220: I think that adding this as a separate step, after `if self.count: ... else: ...` would be more readable. It's a separate list comprehension then, and a bit more explicit. This is a kind of postprocessing of count values, so I think that it makes sense. Also warning about scaling for bit variant would be then closer to calling `_scale_by_hac()`

skfp/fingerprints/atom_pair.py
```diff
@@ -206,14 +215,31 @@ def _calculate_fingerprint(
...
+            if self.normalize:
+                warnings.warn("Scaling by HAC can only be applied to count vectors.")
 
         return csr_array(X) if self.sparse else np.array(X)
+
+    def _scale_by_hac(self, fingerprint: np.ndarray, mol: Mol) -> np.ndarray:
+        """Scale computed fingerprint by the heavy atom count (HAC).
+
+        Vector values are expressed in percent and rounded to the nearest integer.
+        """
```
j-adamczyk, line 244: If this is a private method, maybe just leave a regular comment? Nobody should be interested in the docstring anyway.
```
# scale values to percentages, rounded to the nearest integer
```

tests/fingerprints/atom_pair.py
```diff
@@ -45,6 +45,23 @@ def test_atom_pair_count_fingerprint(smiles_list, mols_list):
     assert X_skfp.shape == (len(smiles_list), atom_pair_fp.fp_size)
 
 
+def test_atom_pair_count_fingerprint_normalized(smiles_list, mols_list):
```
j-adamczyk, line 48: Maybe `hac_scaled` instead of `normalized`, to be more specific?

tests/fingerprints/atom_pair.py
```diff
@@ -45,6 +45,23 @@ def test_atom_pair_count_fingerprint(smiles_list, mols_list):
...
+    atom_pair_fp = AtomPairFingerprint(
+        sparse=False, count=True, normalize=True, n_jobs=-1
+    )
+    X_skfp = atom_pair_fp.transform(smiles_list)
+
+    fp_gen = GetAtomPairGenerator()
+    X_rdkit = np.array([fp_gen.GetCountFingerprintAsNumPy(mol) for mol in mols_list])
+    X_rdkit_scaled = [
+        np.rint(100 * fp / mol.GetNumHeavyAtoms())
```
j-adamczyk, line 57: Why `rint` here and `np.round` in `_scale_by_hac`?

j-adamczyk (changes requested):

tests/fingerprints/atom_pair.py
BlazejNowicki: Previously I used `rint` in `_scale_by_hac`. It's fixed now with `round` in both.

skfp/fingerprints/atom_pair.py
BlazejNowicki: Minor drawback is that we have to abandon convention to always operate on X variable but I assume it's still worth it.

skfp/fingerprints/atom_pair.py
BlazejNowicki: Agreed on plain `[2]` at the end of the description

skfp/fingerprints/atom_pair.py
```diff
@@ -206,14 +216,24 @@ def _calculate_fingerprint(
             countSimulation=self.count_simulation,
         )
         if self.count:
-            X = [
+            Y = [
```
j-adamczyk, line 219: Why Y? X was good

skfp/fingerprints/atom_pair.py
```diff
@@ -206,14 +216,24 @@ def _calculate_fingerprint(
...
                 gen.GetFingerprintAsNumPy(mol, confId=conf_id)
                 for mol, conf_id in zip(X, conf_ids)
             ]
 
-        return csr_array(X) if self.sparse else np.array(X)
+        if self.normalize:
+            if self.count:
+                Y = [self._scale_by_hac(fp, mol) for fp, mol in zip(Y, X)]
+            else:
+                warnings.warn("Scaling by HAC can only be applied to count vectors.")
```
j-adamczyk, line 233: Maybe add "No HAC scaling will be applied" as a second sentence, to be more explicit?

skfp/fingerprints/atom_pair.py
```diff
@@ -63,6 +64,11 @@ class AtomPairFingerprint(FingerprintTransformer):
     count : bool, default=False
         Whether to return binary (bit) features, or their counts.
 
+    normalize: bool, default=False
```
j-adamczyk, line 67: I would change this to `scale_by_hac`, to be more precize. Molecule normalization also exists, and is something else

j-adamczyk (changes requested):

skfp/fingerprints/atom_pair.py
BlazejNowicki: X is a list of molecules and I can't overwrite it with fingerprints because I need list of molecules to perform scaling

j-adamczyk: Ah, I see. So maybe `fps` variable, as a abbreviation for `fingerprints`? I think it's quite commonly used

BlazejNowicki: That was the advantage of doing it with one list comprehension as it was before

j-adamczyk (approved):

mjste (approved):

my-alaska (approved):

PR #126, opened 2024-04-10
Avalon docs

mjste:


skfp/fingerprints/avalon.py
```diff
@@ -12,7 +12,64 @@
 
 
 class AvalonFingerprint(FingerprintTransformer):
-    """Avalon fingerprint."""
+    """
+    Avalon fingerprint.
+
+    The implementation uses RDKit [1]_. This is a hashed fingerprint, where
```
j-adamczyk, line 18: Generally, we don't cite RDKit, since it is quite obvious. Also, we would have to cite it basically everywhere.

skfp/fingerprints/avalon.py
```diff
@@ -12,7 +12,64 @@
...
+    fragments are computed based on atom environments. Avalon is a path-based
+    fingerprint.
```
j-adamczyk, line 20: I would say that fragments are based on multiple feature classes, including e.g. atom and bond type counts, rings and paths. In general, Avalon has quite a large collection of different fragments, and even their combinations.

skfp/fingerprints/avalon.py
j-adamczyk: Also, maybe "uses RDKit and Avalon toolkit", since RDKit really is a wrapper for it?

skfp/fingerprints/avalon.py
j-adamczyk: Maybe add "The detailed descriptions of feature classes are in the supplementary material of [3]_", and this will replace the "The fingerprint is based..."

skfp/fingerprints/avalon.py
```diff
@@ -12,7 +12,64 @@
...
+    n_jobs : int, default=None
+        Number of parallel jobs. If -1, then the number of jobs is set to the number of CPU cores.
+
+    verbose : int, default=0
+        Verbosity level.
+
+    Attributes
+    ----------
+    n_features_out : int
+        Number of output features.
```
j-adamczyk, line 46: Add:
```
Equal to `fp_size`.
```

skfp/fingerprints/avalon.py
```diff
@@ -12,7 +12,64 @@
...
+    References
+    ----------
+    .. [1] RDKit implementation of Avalon fingerprints
+        https://www.rdkit.org/docs/source/rdkit.Avalon.pyAvalonTools.html
+    .. [2] Avalon toolkit
+        https://sourceforge.net/projects/avalontoolkit/
+    .. [3] Gedeck, P., Rohde, B., & Bartels, C. (2006).
+        "QSAR − How Good Is It in Practice? Comparison of Descriptor Sets on an Unbiased
+        Cross Section of Corporate Data Sets."
+        Journal of Chemical Information and Modeling, 46(5), 1924–1936. doi:10.1021/ci050413p
```
j-adamczyk, line 57: Use reference format like in Atom Pairs. We should stick to one formatting. It will also create nice hyperlinks. For authors, use MLA format instead of APA (first one on Google Scholar).
```
    .. [3] `Gedeck, Peter, Bernhard Rohde, and Christian Bartels
        "QSAR − How Good Is It in Practice? Comparison of Descriptor Sets on an Unbiased
        Cross Section of Corporate Data Sets."
        J. Chem. Inf. Model. 2006, 46, 5, 1924–1936`_
```

j-adamczyk (changes requested):

skfp/fingerprints/avalon.py
mjste: Yes it is. Painfully checked it

j-adamczyk (approved):

PR #129, opened 2024-04-10
Fix + refactor example_classification.py

mjste:


benchmarking/example_classification.py
```diff
@@ -10,33 +11,37 @@
...
+    ("ogbg-molbbbp", "p_np"),
 ]
 
-for dataset_name, property_name in zip(dataset_names, property_names):
+
+classifier_parameters = [
+    (RandomForestClassifier, {"n_jobs": -1, "class_weight": "balanced"}),
+    (
+        LogisticRegression,
+        {"class_weight": "balanced", "penalty": None, "n_jobs": -1},
```
j-adamczyk, line 34: I would use `LogisticRegressionCV` here, with and default tuning of L2 penalty. Linear classifiers are sensitive to regularization strength

benchmarking/example_classification.py
```diff
@@ -10,33 +11,37 @@
...
+
+for dataset_name, property_name in dataset_params:
     dataset = GraphPropPredDataset(name=dataset_name, root="../dataset")
     split_idx = dataset.get_idx_split()
-    train_idx = np.array(split_idx["train"])
-    valid_idx = np.array(split_idx["valid"])
-    test_idx = np.array(split_idx["test"])
+    train_idx = np.array(split_idx["train"])[:SIZE]
+    valid_idx = np.array(split_idx["valid"])[:SIZE]
+    test_idx = np.array(split_idx["test"])[:SIZE]
```
j-adamczyk, line 44: I think that we shouldn't limit the sizes here. Obviously, some fingerprints will be quite slow, but if we want to get classification results with them, this is the price.

mjste: That is our goal in the end. However, testing it with >100 molecules already takes a few minutes. This was just for having shorted feedback loop if it even works

benchmarking/example_classification.py
mjste: We can change `LogisticRegression` to `LogisticRegressionCV`. About L2: we already use it - it's default.

benchmarking/example_classification.py
```diff
@@ -54,54 +66,45 @@
...
+        except RuntimeError as e:
+            print(f" - Error: {e}")
+            records[-1]["error"] = str(e)
+            continue
+        except ValueError as e:
+            try:
+                conf_gen = ConformerGenerator()
+                X_train_conf = conf_gen.transform(X_train_mols)
+                X_valid_conf = conf_gen.transform(X_valid_mols)
+                X_test_conf = conf_gen.transform(X_test_mols)
```
j-adamczyk, line 98: We should generate conformers once and then use them as needed. Recomputing them will be very expensive. Maybe we could introduce an attribute to indicate which fingerprints require conformers, that would get rid of this try/except

benchmarking/example_classification.py
```diff
@@ -54,54 +66,45 @@
...
+                X_fp_valid = fp_transformer.transform(X_valid_conf)
+                X_fp_test = fp_transformer.transform(X_test_conf)
+            except ValueError as e:
+                print(e.__class__.__name__)
+                print(f" - Error: {e}")
+                records[-1]["error"] = str(e)
+                continue
 
         scaler = MinMaxScaler()
         X_fp_train = scaler.fit_transform(X_fp_train)
```
j-adamczyk, line 110: 1. Scaling should be done after imputation
2. We can have extremely high values for some descriptors. While they can be clipped, this should be addressed here. Maybe use RobustScaler instead of MinMaxScaler?
3. Replace second `fit_transform` with `transform`

benchmarking/example_classification.py
```diff
@@ -112,18 +115,30 @@
...
         print(f" - Time of fingerprints computing : {round(execution_time,2)}s")
         records[-1]["execution_time"] = execution_time
-        for classifier, clf_name, clf_kwargs in zip(
-            classifiers, clf_names, classifier_kwargs
-        ):
+
+        imputer = SimpleImputer(strategy="constant", fill_value=0)
+        X_fp_train = imputer.fit_transform(X_fp_train)
+        X_fp_valid = imputer.transform(X_fp_valid)
+        X_fp_test = imputer.transform(X_fp_test)
```
j-adamczyk, line 122: Technically, this will not work properly for descriptors, just for fingerprints, which should never have missing values anyway. Maybe replace with median imputation?

j-adamczyk (changes requested):

benchmarking/example_classification.py
mjste: The attribute would be very helpful

benchmarking/example_classification.py
mjste: Alright

benchmarking/example_classification.py
mjste: Sure

benchmarking/example_classification.py
mjste: Done

benchmarking/example_classification.py
```diff
@@ -54,76 +71,88 @@
...
+
+    try:
+        conf_gen = ConformerGenerator()
+        X_train_conf = conf_gen.transform(X_train_mols)
+        X_valid_conf = conf_gen.transform(X_valid_mols)
+        X_test_conf = conf_gen.transform(X_test_mols)
+        conformers_generated = True
+    except ValueError as e:
+        conformers_generated = False
+        print(f" - Error: {e}")
```
j-adamczyk, line 86: If we can't generate conformers, we should probably just raise an error and stop

benchmarking/example_classification.py
```diff
@@ -54,76 +71,88 @@
...
+                    X_fp_train = fp_transformer.transform(X_train_conf)
+                    X_fp_valid = fp_transformer.transform(X_valid_conf)
+                    X_fp_test = fp_transformer.transform(X_test_conf)
+                except ValueError as e:
+                    print(e.__class__.__name__)
+                    print(f" - Error: {e}")
+                    records[-1]["error"] = str(e)
+                    continue
+            else:
+                continue
```
j-adamczyk, line 121: You can just always pass molecules with conformers

benchmarking/example_classification.py
```diff
@@ -54,76 +71,88 @@
...
+
+        imputer = SimpleImputer(strategy="median")
+        X_fp_train = imputer.fit_transform(X_fp_train)
+        X_fp_valid = imputer.transform(X_fp_valid)
+        X_fp_test = imputer.transform(X_fp_test)
+
+        scaler = RobustScaler()
+        X_fp_train = scaler.fit_transform(X_fp_train)
+        X_fp_valid = scaler.transform(X_fp_valid)
+        X_fp_test = scaler.transform(X_fp_test)
```
j-adamczyk, line 136: I think we should have explicitly do this only for descriptors, since for fingerprints `RobustScaler` (or basically any scaling) would destroy sparsity. Descriptors list:
```
[
AutocorrFingerprint,
GETAWAYFingerprint,
MordredFingerprint,
MORSEFingerprint,
RDFFingerprint,
WHIMFingerprint
]
```

j-adamczyk (changes requested):

benchmarking/example_classification.py
mjste: how about using ConformerGenerator.transform_x_y ?

benchmarking/example_classification.py
mjste: by _this_ you mean imputing and scaling or only scaling?

j-adamczyk (approved):

PR #133, opened 2024-04-12
Saving Fingerprints

my-alaska:
Add a script for saving fingerprints to .npy files for hiv, BACE and BBBP


benchmarking/saving_fingerprints_to_npy.py
```diff
@@ -0,0 +1,165 @@
...
+    "ogbg-molbbbp",
+]
+
+property_names = [
+    "HIV_active",
+    "Class",
+    "p_np",
+]
+
+fp_names = [
```
j-adamczyk, line 27: A few naming conventions and fixing typos:
- Autocorr -> Autocorrelation (note that we have 2 variants, 2D and 3D there)
- ERG -> ErG
- Getway -> GETAWAY
- MORDRED -> Mordred
- MORSE -> MoRSE
- Phsiochemical-Properties -> Physiochemical Properties
- Topological-Torsion -> Topological Torsion
- Whim -> WHIM

mjste (approved):

benchmarking/saving_fingerprints_to_npy.py
mjste: I would suggest using class.__name__ instead. It will simplify reading saved files and eliminate errors such as typos.

mjste (changes requested):
Fp names need to be fixed

benchmarking/saving_fingerprints_to_npy.py
j-adamczyk: Agreed

j-adamczyk (changes requested):

mjste (approved):

mjste (approved):

PR #149, opened 2024-04-23
Lingo fingerprint

mjste:


skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+
+from skfp.fingerprints.base import FingerprintTransformer
+from skfp.validators import ensure_smiles
+
+
+class LingoFingerprint(FingerprintTransformer):
+    """
+    Lingo fingerprint.
+
+    The Lingo fingerprint [1]_ is a hashed fingerprint based on cdk that counts/flags the occurrences of
```
j-adamczyk, line 21: Remove `based on cdk` - we reimplement this entirely, and this is based basically only on the original paper

j-adamczyk: Maybe `that check the occurrences of substrings...`? Count version is covered by parameter description

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+    substrings of a given length in a SMILES string. The hash function used in our
+    implementation is SHA-1. The Lingo fingerprint is a binary fingerprint by default,
+    but it can also be a count fingerprint.
```
j-adamczyk, line 24: Every fingerprint is binary by default, we don't need to specify this here

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+
+    If option `count` is set to True, the fingerprint vector will contain counts of
+    substrings, otherwise it will contain binary flags.
```
j-adamczyk, line 27: Put this in parameter description, not here

skfp/fingerprints/lingo.py
j-adamczyk: I would more explicitly underline here that while original paper uses raw dictionaries, we return constant-length, hashed fingerprints

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+    ----------
+    fp_size : int, default=1024
+        Size of output vectors, i.e. number of bits for each fingerprint. Must be
+        positive.
+
+    substring_length : int, default=4
+        Length of the substrings to count in the SMILES strings. Must be positive.
+
+    count : bool, default=False
+        Whether to use count fingerprint instead of binary fingerprint.
```
j-adamczyk, line 42: Keep description uniform with other fingerprints, e.g. Atom Pair:
```
    count : bool, default=False
        Whether to return binary (bit) features, or their counts.
```

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+
+    The Lingo fingerprint [1]_ is a hashed fingerprint based on cdk that counts/flags the occurrences of
+    substrings of a given length in a SMILES string. The hash function used in our
+    implementation is SHA-1. The Lingo fingerprint is a binary fingerprint by default,
+    but it can also be a count fingerprint.
+
+    If option `count` is set to True, the fingerprint vector will contain counts of
+    substrings, otherwise it will contain binary flags.
+
+    You can use method :meth:`smiles_to_dicts` to convert SMILES strings to dictionaries of substring,
```
j-adamczyk, line 29: We shouldn't go over 88 characters also in docstrings

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+
+    substring_length : int, default=4
+        Length of the substrings to count in the SMILES strings. Must be positive.
+
+    count : bool, default=False
+        Whether to use count fingerprint instead of binary fingerprint.
+
+    sparse : bool, default=False
+        Whether to return sparse matrix. If True, the output will be a Scipy CSR matrix, otherwise
+        it will be a Numpy array.
```
j-adamczyk, line 46: Again, keep description uniform with other fingerprints:
```
    sparse : bool, default=False
        Whether to return dense NumPy array, or sparse SciPy CSR array.
```

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+        super().__init__(
+            n_features_out=fp_size,
+            sparse=sparse,
+            n_jobs=n_jobs,
+            batch_size=batch_size,
+            verbose=verbose,
+        )
+        self.fp_size = fp_size
+        self.substring_length = substring_length
+        self.count = count
```
j-adamczyk, line 112: Pass this to `super().__init__()`

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Returns
+        -------
+        result: list[defaultdict[str, int]]
+            List of dictionaries containing substring counts.
+        """
+        X = ensure_smiles(X)
+        X = [re.sub(r"[123456789]", "0", x) for x in X]
+        X = [re.sub(r"Cl", "L", x) for x in X]
+        X = [re.sub(r"Br", "R", x) for x in X]
```
j-adamczyk, line 154: Add comment why we're doing this

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+
+        result = []
+        for smiles in X:
+            result_dict: defaultdict[str, int] = defaultdict(int)
+            for i in range(len(smiles) - self.substring_length + 1):
+                result_dict[smiles[i : i + self.substring_length]] += 1
+            result.append(result_dict)
```
j-adamczyk, line 161: Consider an alternative solution:
```
substrings = [smiles[i : i + self.substring_length] for i in range(len(smiles) - self.substring_length + 1)]
substring_counts = dict(Counter(substrings))
```
Which one to use should be based on speed

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+        X = self.smiles_to_dicts(X)
+        X = self._dicts_to_array(X)
+        return csr_array(X) if self.sparse else X
+
+    def _dicts_to_array(self, X: list[defaultdict[str, int]]) -> np.ndarray:
+        dtype = np.uint32 if self.count else np.uint8
+        result = np.zeros((len(X), self.fp_size), dtype=dtype)
+        for i, dictionary in enumerate(X):
+            for key, value in dictionary.items():
+                string_bytes = key.encode()
```
j-adamczyk, line 176: Specify "utf-8" explicitly. Doesn't `bandit` pre-commit hook complain about this?

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,183 @@
...
+                hash_bytes = hashlib.sha1(string_bytes, usedforsecurity=False).digest()
+                hash_index = int.from_bytes(hash_bytes, byteorder="big") % self.fp_size
+                if self.count:
+                    result[i, hash_index] += value
+                else:
+                    result[i, hash_index] = 1
```
j-adamczyk, line 182: First construct the count array, then binarize if needed. if/else in the loop is most likely slower than vectorized NumPy operation.

tests/fingerprints/lingo.py
```diff
@@ -0,0 +1,101 @@
...
+    X_skfp = lingo_fp.transform(smiles)
+
+    expected_array = load_npz(
+        os.path.join("tests", "fingerprints", "lingo_count_sparse.npz")
+    )
+
+    assert np.array_equal(X_skfp.data, expected_array.data)
+    assert X_skfp.shape == (2, 1024)
+    assert X_skfp.dtype == np.uint32
+    assert np.all(X_skfp.data >= 0)
```
j-adamczyk, line 101: I would like to avoid files in tests. They are not human-readable, and those tests rely on the assumption that those files contain ground truth data. I would just compare the parallel and sequential versions like other fingerprints.

j-adamczyk (changes requested):

skfp/fingerprints/lingo.py
mjste: python3 default encoding is UTF* as in [docs](https://docs.python.org/3/howto/unicode.html#the-string-type). Of course I can write it explicitly. Tests have passed, so no - bandit didn't complain.

tests/fingerprints/lingo.py
mjste: It's true that we assume the files are correct. However, without these tests, we are not sure if some changes in code changed output. Comparing sequential processing to parrales does not guarantee if the transformation itself is correct, or at least not changed

skfp/fingerprints/lingo.py
mjste: I have done performance tests on both solutions. I do not see a significant gain (or any difference) from changing solutions. Both solutions transform hiv dataset locally in ~3.7-3.9s.

mjste: I would leave the original solution. I think it's easier to read

skfp/fingerprints/lingo.py
mjste: replacing given fragment with:
```             
                # if self.count:
                #     result[i, hash_index] += value
                # else:
                #     result[i, hash_index] = 1
                result[i, hash_index] += value

        # binarize the result
        result[result > 1] = 1
```
actually increased calculation time for hiv dataset from 3.8s to 5.6s. If you have any more specific recommendation, we can use it. My implementation of binarization is significantly slower. Maybe other changes can help in speeding up the processing.

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class LingoFingerprint(FingerprintTransformer):
+    """
+    Lingo fingerprint.
+
+    The Lingo fingerprint [1]_ is a hashed fingerprint that checks the occurrences of
+    substrings of a given length in a SMILES string. Original paper uses raw dictionaries os output.
+    Our implementation returns constant-length hashed fingerprints. The hash function used in our
+    implementation is SHA-1.
```
j-adamczyk: Do not use first person: `Original paper uses raw dictionaries as output, but here constant-length hashed fingerprints are returned`. Also, `os` is a typo.

Second sentence can he shorter: `SHA-1 hash function is used here`.

skfp/fingerprints/lingo.py
```diff
@@ -0,0 +1,179 @@
...
+from skfp.fingerprints.base import FingerprintTransformer
+from skfp.validators import ensure_smiles
+
+
+class LingoFingerprint(FingerprintTransformer):
+    """
+    Lingo fingerprint.
+
+    The Lingo fingerprint [1]_ is a hashed fingerprint that checks the occurrences of
+    substrings of a given length in a SMILES string. Original paper uses raw dictionaries os output.
```
j-adamczyk: Maybe specify that those substrings are overlapping

skfp/fingerprints/lingo.py
j-adamczyk: Ok

tests/fingerprints/lingo.py
j-adamczyk: Ok, but move test files to a subdirectory, e.g. `data`

tests/fingerprints/lingo.py
```diff
@@ -0,0 +1,101 @@
...
+    X_skfp = lingo_fp.transform(smiles)
+
+    expected_array = load_npz(
+        os.path.join("tests", "fingerprints", "lingo_bit_sparse.npz")
+    )
+
+    assert np.array_equal(X_skfp.data, expected_array.data)
+    assert X_skfp.shape == (2, 1024)
+    assert X_skfp.dtype == np.uint8
+    assert np.all(np.isin(X_skfp.data, [0, 1]))
```
j-adamczyk, line 86: For bit sparse, `X_skfp.data` should be all ones, i.e. `np.all(X_skfp.data == 1)`

tests/fingerprints/lingo.py
j-adamczyk, line 101: This should be `>0`

j-adamczyk (changes requested):

j-adamczyk (approved):

PR #150, opened 2024-04-23
Allow running tests manually

mjste:


mjste:
Why I needed to exclude the test:
https://github.com/actions/setup-python/issues/696#issuecomment-1637587760

PR #204, opened 2024-07-15
Implemented scaffold split and randomized scaffold split

StaniszewskiA:


skfp/utils/splitters.py
```diff
@@ -0,0 +1,76 @@
...
+    test_ids: list[int] = []
+
+    for scaffold_set in scaffold_sets:
+        if len(train_ids) + len(scaffold_set) > train_cutoff:
+            if len(train_ids) + len(valid_ids) + len(scaffold_set) > valid_cutoff:
+                test_ids += scaffold_set
+            else:
+                valid_ids += scaffold_set
+        else:
+            train_ids += scaffold_set
```
j-adamczyk, line 74: See comment above about inverting the list order

j-adamczyk (changes requested):

skfp/utils/splitters.py
```diff
@@ -0,0 +1,76 @@
+from collections.abc import Sequence
+import numpy as np
+from typing import Optional, Any
+
+from rdkit import Chem
+from rdkit.Chem.Scaffolds import MurckoScaffold
+
+from validators import ensure_mols
+
```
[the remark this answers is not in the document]
StaniszewskiA: Could you please specify?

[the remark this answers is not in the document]
StaniszewskiA: Nevermind, I think I see what you mean now.

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+        return train_ids, valid_ids, test_ids
+    else:
+        return (
+            _get_data_from_indices(data, scaffolds, train_ids),
+            _get_data_from_indices(data, scaffolds, valid_ids),
+            _get_data_from_indices(data, scaffolds, test_ids),
+        )
+
+
+def _get_data_from_indices(data, scaffolds, indices):
```
j-adamczyk, line 217: Add typing

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+    return scaffolds
+
+
+def scaffold_train_test_split(
+    data: Sequence[Union[str, Mol]],
+    train_size: float = 0.8,
+    test_size: float = 0.2,
+    return_indices: bool = False,
+    include_chirality: bool = False,
+) -> tuple[list, list]:
```
j-adamczyk, line 55: This should be:
```
Union[tuple[list[Union[str, Mol]], list[Union[str, Mol]], tuple[list[int], list[int]]]
```

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+            train_ids.extend(scaffold_set)
+        else:
+            test_ids.extend(scaffold_set)
+
+    if return_indices:
+        return train_ids, test_ids
+    else:
+        return _get_data_from_indices(
+            data, scaffolds, train_ids
+        ), _get_data_from_indices(data, scaffolds, test_ids)
```
j-adamczyk, line 122: Create two variables first and then return

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def scaffold_train_valid_test_split(
+    data: Sequence[Union[str, Mol]],
+    train_size: float = 0.8,
+    test_size: float = 0.1,
+    valid_size: float = 0.1,
+    return_indices: bool = False,
+    include_chirality: bool = False,
+) -> tuple[list, list, list]:
```
j-adamczyk, line 132: Specify types as above

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+    Split a list of SMILES or RDKit Mol objects into train, validation,
+    and test subsets using Bemis-Murcko scaffolds.
```
j-adamczyk, line 135: You can make those lines longer, I think. We generally try to get ~88 line length

j-adamczyk: Also, add citations and fixes as in the comment above for train-test function

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+
+    Notes
+    -----
+    - This implementation uses Bemis-Murcko scaffolds (Bemis and Murcko, 1996) to group molecules.
+    - The split is fully deterministic, with the smallest scaffold sets assigned to the test subset,
+    larger sets to the validation subset,
+      and the largest sets to the training subset.
+    - The test and validation sets can be slightly larger than specified,
+    ensuring they are at least the desired fraction, but as close as possible in size.
+    - The split fractions (frac_train, frac_valid, frac_test) must sum to 1.
```
j-adamczyk, line 178: Make sure that those notes are identical for both train-test and train-valid-test functions

skfp/utils/splitters.py
```diff
@@ -0,0 +1,221 @@
...
+            _get_data_from_indices(data, scaffolds, valid_ids),
+            _get_data_from_indices(data, scaffolds, test_ids),
+        )
+
+
+def _get_data_from_indices(data, scaffolds, indices):
+    """
+    Helper function to retrieve data from indices.
+    """
+    return [data[i] if isinstance(data[0], str) else scaffolds[i] for i in indices]
```
j-adamczyk, line 221: 1. This will be really slow for lists. Change `indices` to a set.
2. Why `isinstance(data[0], str)`?

j-adamczyk (changes requested):

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    include_chirality: bool = False,
+) -> Union[
+    tuple[list[str], list[str]],
+    tuple[list[Mol], list[Mol]],
+    tuple[list[int], list[int]],
+]:
+    """
+    Split a list of SMILES or RDKit 'Mol' objects into train and test subsets using Bemis-Murcko [1]_ scaffolds.
+
+    This function ensures that similar molecules (sharing the same scaffold) are in the same split.
```
j-adamczyk, line 26: Maybe focus more on the fact that train and test are structurally different? I mean, this is kind of the same, but the impact of the wording should stress that by using scaffold split, we strive for generalization on structurally novel molecules

j-adamczyk: Also, MoleculeNet introduced scaffold split as an approximation to the time split, assuming that new molecules (test set) will be indeed structurally different in terms of scaffolds from the training set. Maybe mention this?

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    This method, suggested in the MoleculeNet [2]_ paper, helps to test the model's ability to
+    generalize to entirely new scaffolds.
+
+    The split is fully deterministic, with the smallest scaffold sets assigned to the test
+    subset and the largest to the training subset.
```
j-adamczyk, line 31: Maybe "and the rest to the training subset"? Because we can have kind of "average-sized" scaffold sets, which also get put into the training subset

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+        - if `return_indices` is False, returns lists of SMILES strings
+        or RDKit 'Mol' objects depending on the input
+        - if `return_indices` is True, returns lists of indices
+
+    References
+    ----------
+    .. [1] Bemis, G. W., & Murcko, M. A. (1996). The properties of known drugs. 1. Molecular frameworks.
+    Journal of Medicinal Chemistry, 39(15), 2887-2893.
+    .. [2] Z. Wu, B. Ramsundar, E. N. Feinberg, J. Gomes, C. Geniesse, A. S. Pappu, K. Leswing, V. Pande (2017).
+    MoleculeNet: A Benchmark for Molecular Machine Learning. Chemical Science, 9(2), 513-530.
```
j-adamczyk, line 66: Use multiline format like in fingerprints, it makes it easier to manage docs. Also include backticks and URL to the paper, and put the name of the publication in quotes. Example:
```
.. [1] `John Smith
    "Lorem Ipsum"
    Chemical Science, 9(2), 513-530.
    <url_to_paper_page>`_
```
Note backticks and underscore - this makes this a clickable link in the resulting docs

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+            train_ids.extend(scaffold_set)
+        else:
+            test_ids.extend(scaffold_set)
+
+    if return_indices:
+        return train_ids, test_ids
+    else:
+        return _get_data_from_indices(
+            data, scaffolds, train_ids
+        ), _get_data_from_indices(data, scaffolds, test_ids)
```
j-adamczyk, line 88: Explicitly create two variables and return them, it will be more readable

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    valid_size: float = 0.1,
+    return_indices: bool = False,
+    include_chirality: bool = False,
+) -> Union[
+    tuple[list[str], list[str], list[str]],
+    tuple[list[Mol], list[Mol], list[Mol]],
+    tuple[list[int], list[int], list[int]],
+]:
+    """
+    Split a list of SMILES or RDKit 'Mol' objects into train, validation,
```
j-adamczyk, line 104: I won't comment here on the description - introduce changes in train-test split above, and copy the result here, and add the appropriate changes.

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+
+    train_cutoff: int = int(train_size * len(data))
+    train_ids: list[int] = []
+    test_ids: list[int] = []
+
+    for scaffold_set in scaffold_sets:
+        if len(train_ids) < train_cutoff:
+            train_ids.extend(scaffold_set)
+        else:
+            test_ids.extend(scaffold_set)
```
j-adamczyk, line 81: You have the scaffold sets sorted from smallest to largest. So I would write this as:
```
desired_test_size = int(test_size * len(data))

for scaffold_set in scaffold_sets:
    if len(test_ids) < desired_test_size:
        test_ids.extend(scaffold_set)
    else:
        train_ids.extend(scaffold_set)
```

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    valid_ids: list[int] = []
+    test_ids: list[int] = []
+
+    for scaffold_set in scaffold_sets:
+        if len(train_ids) < train_cutoff:
+            train_ids.extend(scaffold_set)
+        elif len(valid_ids) < valid_cutoff:
+            valid_ids.extend(scaffold_set)
+        else:
+            test_ids.extend(scaffold_set)
```
j-adamczyk, line 171: See my comment above about train-test set division. Here, I would use the order test -> valid -> rest into train

skfp/utils/splitters.py
j-adamczyk: This would also follow the descriptions from papers more explicitly - we *first* construct the test set from smallest scaffold sets, and *then* put the rest into train set

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    train_cutoff: int = int(train_size * len(data))
+    train_ids: list[int] = []
+    test_ids: list[int] = []
+
+    for scaffold_set in scaffold_sets:
+        if len(train_ids) < train_cutoff:
+            train_ids.extend(scaffold_set)
+        else:
+            test_ids.extend(scaffold_set)
+
```
j-adamczyk, line 82: Check here if train or test is empty, if so, throw an error

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    test_ids: list[int] = []
+
+    for scaffold_set in scaffold_sets:
+        if len(train_ids) < train_cutoff:
+            train_ids.extend(scaffold_set)
+        elif len(valid_ids) < valid_cutoff:
+            valid_ids.extend(scaffold_set)
+        else:
+            test_ids.extend(scaffold_set)
+
```
j-adamczyk, line 172: Check here if train/valid/test is empty. For empty train or test throw error, and for empty valid print warning and suggest using `scaffold_train_test_split` instead

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    Returns
+    -------
+    scaffolds : dict
+        A dictionary where keys are Bemis-Murcko scaffolds (as SMILES strings) and
+        values are lists of indices pointing to molecules sharing the same scaffold.
+
+    Notes
+    -----
+    - This implementation uses Bemis-Murcko scaffolds (Bemis and Murcko, 1996) to group molecules.
+    - Each scaffold is represented as a SMILES string.
```
j-adamczyk, line 206: Move to main description above

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+    result = []
+    index_set = set(indices)
+
+    for i in index_set:
+        if isinstance(data[i], str):
+            item = data[i]
+        elif isinstance(data[i], Mol):
+            item = data[i]
+        else:
+            raise ValueError("Unsupported data type in input.")
```
j-adamczyk, line 235: The type doesn't really matter here, you are just taking the indexes. This can be simplified to:
```
indices = set(indices)
return [data[idx] for idx in indices]
```

skfp/utils/splitters.py
```diff
@@ -0,0 +1,239 @@
...
+        scaffold = MurckoScaffold.MurckoScaffoldSmiles(
+            mol=molecule, includeChirality=include_chirality
+        )
+        scaffolds[scaffold].append(ind)
+
+    return scaffolds
+
+
+def _get_data_from_indices(
+    data: Sequence[Union[str, Mol]], scaffolds: dict[str, list], indices: Sequence[int]
```
j-adamczyk, line 221: You don't need scaffold here, do you?

j-adamczyk (changes requested):

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+    indices : sequence
+        List of indices to retrieve from `data`.
+
+    Returns
+    -------
+    list
+        List of data elements at the specified indices.
+    """
+    indices = set(indices)
+    return [data[idx] for idx in indices]
```
j-adamczyk, line 463: Can be a 1-liner

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+    Returns
+    ----------
+    subsets : tuple[list, list]
+        A tuple of train and test subsets. The format depends on `return_indices`:
+        - if `return_indices` is False, returns lists of SMILES strings
+        or RDKit `Mol` objects depending on the input
+        - if `return_indices` is True, returns lists of indices
+    additional_data_splitted: list[sequence]:
+        the method may return any additional data splitted in the same way as
+        provided SMILES
```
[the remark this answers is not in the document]
j-adamczyk: This way, we also don't need to specify `additional_data_splitted` separately here

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+    """
+    _validate_split_sizes(train_size=train_size, test_size=test_size)
+    train_size, test_size = _fill_missing_sizes(train_size, test_size)
+    scaffolds = _create_scaffolds(list(data), include_chirality)
+    scaffold_sets = sorted(scaffolds.values(), key=len)
+
+    train_ids: list[int] = []
+    test_ids: list[int] = []
+
+    train_ids, test_ids = _split_ids(scaffold_sets, len(data), test_size)
```
j-adamczyk, line 113: I would definitely prefer to have this here. If this causes cyclomatic complexity hook to fail, raise the file-wide limit in pre-commit hook config

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+    test_subset: list[Any] = []
+
+    if return_indices:
+        train_subset = train_ids
+        test_subset = test_ids
+    else:
+        train_subset = _get_data_from_indices(data, train_ids)
+        test_subset = _get_data_from_indices(data, test_ids)
+
+    _check_subsets(train_subset, test_subset)
```
j-adamczyk, line 125: 1. Change to take one subset and call separately
2. Change name to e.g. `_ensure_nonempty_list`

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+    additional_data_split: list[Sequence[Any]] = []
+
+    additional_data_split = (
+        _split_additional_data(list(additional_data), [train_ids, test_ids])
+        if additional_data
+        else []
+    )
+
+    if additional_data:
+        return train_subset, test_subset, additional_data_split
```
j-adamczyk, line 136: Unpack additional data split, i.e. the returned values should be a flat tuple, without a nested list as a last element

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+        "include_chirality": ["boolean"],
+        "return_indices": ["boolean"],
+    }
+)
+def scaffold_train_valid_test_split(
+    data: Sequence[Union[str, Mol]],
+    *additional_data: list[Sequence],
+    train_size: Optional[float] = None,
+    test_size: Optional[float] = None,
+    valid_size: Optional[float] = None,
```
j-adamczyk, line 169: I would keep the train-valid-test order everywhere

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+        Journal of Medicinal Chemistry, 39(15), 2887-2893.
+        https://www.researchgate.net/publication/14493474_The_Properties_of_Known_Drugs_1_Molecular_Frameworks`_
+
+    .. [2] `Z. Wu, B. Ramsundar, E. N. Feinberg, J. Gomes, C. Geniesse, A. S. Pappu, K. Leswing, V. Pande
+        "MoleculeNet: A Benchmark for Molecular Machine Learning."
+        Chemical Science, 9(2), 513-530.
+        https://www.researchgate.net/publication/314182452_MoleculeNet_A_Benchmark_for_Molecular_Machine_Learning`_
+
+    """
+    _validate_split_sizes(train_size, test_size)
```
j-adamczyk, line 250: This has slightly other logic than for train-valid-test. Probably this function can be removed and put directly here and in the train-test function

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+                "Cannot determine test_size. Provide train_size and valid_size."
+            )
+
+    elif valid_size is None:
+        valid_size = 0.0
+        warnings.warn(
+            "valid_size set to 0, consider using scaffold_train_test_split() instead"
+        )
+
+    return train_size, test_size, valid_size
```
j-adamczyk, line 354: We have quite complex logic here. Maybe simplify this and require user to either pass no of those 3 sizes (and use 80-10-10% then), or pass all three and just check that they sum up to 1, and print warning on zero-size validation size.

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+    valid_size : float or None
+        The proportion of the data to use for validation.
+    test_size : float or None
+        The proportion of the data to use for testing.
+
+    Returns
+    -------
+    tuple
+        A tuple containing the proportions for training, validation, and testing sets.
+    """
```
j-adamczyk, line 319: For internal functions, we don't keep full docstrings, since their maintenance would be too much trouble. Just the main description is enough, without `Parameters` and `Returns`. This applies to all internal functions, with underscore at the beginning.

skfp/utils/validators.py
```diff
@@ -1,10 +1,85 @@
...
+                        raise ValueError(
+                            f"Invalid value for parameter '{param}': {value}"
+                        )
+
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
```
j-adamczyk, line 81: We use scikit-learn for all this, no need to reimplement anything. Just import what's needed, see fingerprint classes for examples.

tests/utils/splitters.py
```diff
@@ -0,0 +1,123 @@
...
+    ]
+
+    return [Chem.MolFromSmiles(smiles) for smiles in monosaccharides_smiles]
+
+
+def test_train_test_split_total_molecule_count(all_molecules):
+    train_split, test_split = scaffold_train_test_split(
+        all_molecules, train_size=0.8, test_size=0.2
+    )
+    assert len(train_split) + len(test_split) == len(all_molecules)
```
j-adamczyk, line 73: We can ensure precise lengths in all tests here, since we control the entire test data. Add `assert` for those.

skfp/utils/splitters.py
```diff
@@ -0,0 +1,630 @@
...
+]:
+    """
+    Split a list of SMILES or RDKit `Mol` objects into train and test subsets using Bemis-Murcko [1]_ scaffolds.
+
+    This function ensures that similar molecules (sharing the same scaffold) are in the same split.
+    This method, suggested in the MoleculeNet [2]_ paper, helps to test the model's ability to
+    generalize to entirely new scaffolds. MoleculeNet introduced the scaffold split as an approximation
+    to the time split, assuming that new molecules (test set) will be structurally different in terms of
+    scaffolds from the training set.
+
```
j-adamczyk, line 54: Maybe add a description about known limitations, e.g. disconnected molecules and those without rings get no scaffold and are put into one group regardless of their structure. Same for train-valid-test docs.

j-adamczyk (changes requested):

skfp/utils/splitters.py
```diff
@@ -0,0 +1,479 @@
...
+    train_size: Optional[float],
+    valid_size: Optional[float],
+    test_size: Optional[float],
+) -> tuple[float, float, float]:
+    """
+    Ensure the sum of train_size, valid_size, and test_size equals 1.0 and provide default values if necessary.
+    """
+    train_size, test_size, valid_size = _calculate_missing_sizes(
+        train_size, test_size, valid_size
+    )
```
j-adamczyk, line 458: Paste that function here. We generally want to avoid functions used only in one place

j-adamczyk (changes requested):
General comment, move files a bit:
- create `skfp/model_selection` package
- put `scaffold_split.py` there with splitting functions
- any common functions, which could also be used by other splits, should be in `skfp/model_selection/utils.py`
- also change tests structure appropriately

skfp/utils/validators.py
```diff
@@ -1,6 +1,8 @@
+import numbers
 from collections.abc import Sequence
-from typing import Any
+from typing import Any, Callable, Union
 
+import numpy as np
```
j-adamczyk, line 5: Are those changes necessary?

j-adamczyk (changes requested):

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,365 @@
...
+from skfp.utils.validators import ensure_mols
+
+
+@validate_params(
+    {
+        "data": ["sequence"],
+        "additional_data": ["list"],
+        "train_size": [
+            Interval(RealNotInt, 0, 1, closed="neither"),
+            Interval(Integral, 1, float("inf"), closed="left"),
```
j-adamczyk, line 29: Does `np.inf` work? It would be a bit better than converting string

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,365 @@
...
+    ensure_nonempty_list(train_subset)
+    ensure_nonempty_list(test_subset)
+
+    additional_data_split: list[Sequence[Any]] = []
+
+    additional_data_split = (
+        split_additional_data(list(additional_data), [train_ids, test_ids])
+        if additional_data
+        else []
+    )
```
j-adamczyk, line 167: Add test for additional data. `make test-coverage` may help, since test coverage highlights which lines aren't tested

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,365 @@
...
+            train_ids.extend(scaffold_set)
+
+    if not train_ids:
+        raise ValueError(
+            "Train set is empty. Adjust the train_size or check provided data."
+        )
+    if not test_ids:
+        raise ValueError(
+            "Test set is empty. Adjust the test_size or check provided data."
+        )
```
j-adamczyk, line 311: Are those checks necessary, if you call `ensure_nonempty_list` below?

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,365 @@
...
+        train_subset = train_ids
+        valid_subset = valid_ids
+        test_subset = test_ids
+    else:
+        train_subset = get_data_from_indices(data, train_ids)
+        valid_subset = get_data_from_indices(data, valid_ids)
+        test_subset = get_data_from_indices(data, test_ids)
+
+    ensure_nonempty_list(train_subset)
+    ensure_nonempty_list(valid_subset)
```
j-adamczyk, line 327: I think this shoule be `test_subset`. Otherwise warning below doesn't really make sense

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,365 @@
...
+) -> dict[str, list]:
+    """
+    Generate Bemis-Murcko scaffolds for a list of SMILES strings or RDKit `Mol` objects.
+    This implementation uses Bemis-Murcko scaffolds to group molecules.
+    Each scaffold is represented as a SMILES string.
+    """
+    scaffolds = defaultdict(list)
+    molecules = ensure_mols(data)
+
+    for ind, molecule in enumerate(molecules):
```
j-adamczyk, line 359: Use `idx, mol` here, it will be more uniform with rest of code

skfp/model_selection/utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+def ensure_nonempty_list(subset: list) -> None:
+    """
+    Check if the provided subset is empty.
+    """
+    if len(subset) == 0:
+        raise ValueError("One of the subsets is empty.")
+
+
+def fill_missing_sizes(
```
j-adamczyk, line 19: Maybe `validate_train_test_sizes`, to keep uniform with train-valid-test variant?

skfp/model_selection/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    train_size: Optional[float], test_size: Optional[float]
+) -> tuple[float, float]:
+    """
+    Fill in missing sizes for train and test sets based on the provided sizes.
+    """
+    if train_size is None and test_size is None:
+        train_size = 0.8
+        test_size = 0.2
```
j-adamczyk, line 27: Maybe just `return 0.8, 0.2`, it will be slightly shorter

skfp/model_selection/utils.py
```diff
@@ -0,0 +1,87 @@
...
+    if train_size is None:
+        if test_size is not None:
+            train_size = 1 - test_size
+        else:
+            raise ValueError("test_size must be provided when train_size is None")
```
j-adamczyk, line 32: This is not true, first `if` covers this case. After `if train_size is None and test_size is None`, you know that at least one of those is not `None`.

j-adamczyk: So this part simplifies to:
```
if train_size is None:
    test_size = 1 - train_size
elif test_size is None:
    train_size = 1 - test_size
else:
    # check that they sum to approximately 1.0
    ...

return train_size, test_size
```

The default case is that both are provided and not `None`, so we can check here if they properly sum up to 1.0

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,365 @@
...
+    if train_size is None and test_size is None:
+        raise ValueError("Either train_size or test_size must be provided")
+    if (
+        train_size is not None
+        and test_size is not None
+        and not np.isclose(train_size + test_size, 1.0)
+    ):
+        raise ValueError("train_size and test_size must sum to 1.0")
+
+    train_size, test_size = fill_missing_sizes(train_size, test_size)
```
j-adamczyk, line 125: 1. First `if` clashes with default values and should be removed
2. Second `if` can be checked inside `fill_missing_sizes`

skfp/model_selection/scaffold_split.py
StaniszewskiA: It does, for some reason I haven't considered

skfp/model_selection/scaffold_split.py
StaniszewskiA: Correct

skfp/model_selection/scaffold_split.py
StaniszewskiA: My mistake; the original scikit-learn implementation uses 'None' here.

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,167 @@
...
+    ]
+
+    return [Chem.MolFromSmiles(smiles) for smiles in all_smiles]
+
+
+@pytest.fixture
+def disconnected_molecules() -> list[str]:
+    disconnected_smiles: list[str] = ["C1CC1.OCC", "C1CCC1.OCC", "CC.OCC"]
+
+    return [Chem.MolFromSmiles(smiles) for smiles in disconnected_smiles]
```
j-adamczyk, line 36: If you are using a list in only one test, just make it there. Fixtures are only for data for multiple tests

j-adamczyk (changes requested):

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,371 @@
...
+    validate_train_test_sizes,
+    validate_train_valid_test_split_sizes,
+)
+from skfp.utils.validators import ensure_mols
+
+
+@validate_params(
+    {
+        "data": ["array-like"],
+        "additional_data": ["tuple"],
```
j-adamczyk, line 24: `tuple` does not work? If so, just use `Any`

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,371 @@
...
+    additional_data_split: list[Sequence[Any]] = (
+        split_additional_data(list(additional_data), train_idxs, test_idxs)
+        if additional_data
+        else []
+    )
+
+    if additional_data:
+        return train_subset, test_subset, additional_data_split
+    else:
+        return train_subset, test_subset
```
j-adamczyk, line 169: Simpler:
```
if additional_data:
    additional_data_split: list[Sequence[Any]] = split_additional_data(list(additional_data), train_idxs, test_idxs)
    return train_subset, test_subset, additional_data_split
else:
    return train_subset, test_subset
```

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+@validate_params(
+    {
+        "data": ["array-like"],
+        "additional_data": ["tuple"],
```
j-adamczyk, line 175: Like above

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,371 @@
...
+    additional_data_split: list[Sequence[Any]] = (
+        split_additional_data(list(additional_data), train_idxs, valid_ids, test_idxs)
+        if additional_data
+        else []
+    )
+
+    if additional_data:
+        return train_subset, valid_subset, test_subset, additional_data_split
+    else:
+        return train_subset, valid_subset, test_subset
```
j-adamczyk, line 346: Like above

skfp/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+def _create_scaffolds(
+    data: Sequence[Union[str, Mol]], include_chirality: bool = False, use_csk=False
```
j-adamczyk, line 350: `use_csk: bool = False`

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,215 @@
...
+
+def test_scaffold_count_for_monosaccharides():
+    smiles = [
+        "O=c1[nH]c(=O)c2[nH]cnc2[nH]1",
+        "Cn1c(=O)c2c(ncn2C)n(C)c1=O",
+        "Cn1cnc2c1c(=O)[nH]c(=O)n2C",
+        "Cn1c(=O)c2[nH]cnc2n(C)c1=O",
+    ]
+    scaffolds = _create_scaffolds(smiles)
+    assert len(scaffolds) == 1
```
j-adamczyk, line 165: 1. Add comments here:
```
    "O=c1[nH]c(=O)c2[nH]cnc2[nH]1",  # xanthine
    "Cn1c(=O)c2c(ncn2C)n(C)c1=O",    # caffeine
    "Cn1cnc2c1c(=O)[nH]c(=O)n2C",    # theobromine
    "Cn1c(=O)c2[nH]cnc2n(C)c1=O",    # theophylline

```
2. Name should be `test_scaffold_count_for_xanthines`

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,215 @@
...
+        "C1CN=C(C2=CC=CC=C2F)N=C1",
+        "C1CN=C(C2=CC=CC=C2Cl)N=C1",
+    ]
+
+    scaffolds = _create_scaffolds(benzodiazepines_smiles)
+    assert len(scaffolds) == 1
+
+
+def test_scaffold_count_for_monosaccharides():
+    smiles = [
```
j-adamczyk, line 158: `smiles_list`?

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,215 @@
...
+    assert len(scaffolds) <= len(all_molecules)
+
+
+def test_no_ring_molecules(no_ring_molecules):
+    scaffolds = _create_scaffolds(no_ring_molecules)
+    assert len(scaffolds) == 1
+
+
+def test_scaffold_count_for_benzodiazepines():
+    benzodiazepines_smiles: list[str] = [
```
j-adamczyk, line 147: `smiles_list` to keep uniformity with test below

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,215 @@
...
+    assert len(test_split) == 1
+
+
+def test_scaffold_creation_total_count(all_molecules):
+    scaffolds = _create_scaffolds(all_molecules)
+    assert len(scaffolds) <= len(all_molecules)
+
+
+def test_no_ring_molecules(no_ring_molecules):
+    scaffolds = _create_scaffolds(no_ring_molecules)
```
j-adamczyk, line 142: You use those molecules only in this one place, change to `smiles_list` and paste here instead of fixture

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,215 @@
...
+    ):
+        _ = scaffold_train_test_split(smiles, use_csk=True)
+
+
+def test_scaffold_train_test_split_returns_molecules(all_molecules):
+    mols = [Chem.MolFromSmiles(smiles) for smiles in all_molecules]
+    train_set, test_set = scaffold_train_test_split(mols, return_indices=False)
+
+    assert all(
+        isinstance(train, Chem.MolFromSmiles("").__class__) for train in train_set
```
j-adamczyk, line 183: Why not `isinstance(train, Mol)`?

j-adamczyk: Also applies to all places below

tests/model_selection/utils.py
```diff
@@ -0,0 +1,133 @@
...
+@pytest.fixture
+def additional_data() -> list[list[Union[str, int, bool]]]:
+    return [["a", "b", "c", "d"], [1, 2, 3, 4], [True, False, True, False]]
+
+
+def test_ensure_nonempty_list_passes():
+    try:
+        ensure_nonempty_list([1, 2, 3])
+    except ValueError as e:
+        pytest.fail(f"Unexpected error: {e}")
```
j-adamczyk, line 29: Why such construction, instead of just letting test fail?

j-adamczyk: Smoke test with just `ensure_nonempty_list([1, 2, 3])` is ok

j-adamczyk (changes requested):

skfp/model_selection/scaffold_split.py
StaniszewskiA: It does not work with either `Any` nor `tuple` for some reason

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -0,0 +1,152 @@
...
+        "C1CN=C(C2=CC=CC=C2)N=C1",
+        "C1CN=C(C2=CC=CC=C2F)N=C1",
+        "C1CN=C(C2=CC=CC=C2Cl)N=C1",
+    ]
+
+    randomized_scaffolds = _create_scaffolds(smiles_list)
+    assert len(randomized_scaffolds) == 1
+
+
+def test_randomized_scaffold_count_for_monosaccharides():
```
j-adamczyk, line 65: Those are xanthines, not monosaccharides

tests/model_selection/scaffold_split.py
```diff
@@ -0,0 +1,208 @@
...
+        "C1CN=C(C2=CC=CC=C2)N=C1",
+        "C1CN=C(C2=CC=CC=C2F)N=C1",
+        "C1CN=C(C2=CC=CC=C2Cl)N=C1",
+    ]
+
+    scaffolds = _create_scaffolds(smiles_list)
+    assert len(scaffolds) == 1
+
+
+def test_scaffold_count_for_monosaccharides():
```
j-adamczyk, line 152: Those are xanthines, not monosaccharides

j-adamczyk (changes requested):

j-adamczyk (changes requested):

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -0,0 +1,380 @@
...
+        ],
+        "test_size": [
+            Interval(RealNotInt, 0, 1, closed="neither"),
+            Interval(Integral, 1, None, closed="left"),
+            None,
+        ],
+        "include_chirality": ["boolean"],
+        "use_csk": ["boolean"],
+        "return_indices": ["boolean"],
+        "random_state": [int, np.random.RandomState],
```
j-adamczyk, line 39: Previous string `"random_state"` was good, do not change it, Scikit-learn validator will work properly

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -0,0 +1,380 @@
...
+)
+def randomized_scaffold_train_test_split(
+    data: Sequence[Union[str, Mol]],
+    *additional_data: Sequence,
+    train_size: Optional[float] = None,
+    test_size: Optional[float] = None,
+    include_chirality: bool = False,
+    use_csk: bool = False,
+    return_indices: bool = False,
+    random_state: Optional[Union[int, np.random.RandomState]] = None,
```
j-adamczyk, line 51: `Optional[Union[int, RandomState, Generator]]`

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -0,0 +1,380 @@
...
+):
+    """
+    Split a list of SMILES or RDKit `Mol` objects into train and test subsets using Bemis-Murcko [1]_ scaffolds.
+
+    This function randomizes molecules within a split split. This method, suggested in the MoleculeNet [2]_ paper,
+    helps to test the model's ability to generalize to entirely new scaffolds. MoleculeNet introduced
+    the scaffold split as an approximation to the time split.
+
+    The `use_csk` parameter allows to choose between using the core structure scaffold (which includes atom types)
+    and the skeleton scaffold (which does not). [3]_
```
j-adamczyk, line 61: Citation should be before the dot, i.e. `(which does not) [3]_.`

j-adamczyk (approved):

PR #234, opened 2024-09-02
Fixed empty set in scaffold split

StaniszewskiA:


tests/model_selection/scaffold_split.py
```diff
@@ -206,3 +206,23 @@ def test_scaffold_train_valid_test_split_return_indices(all_molecules):
...
+
+
+def test_empty_train_subset_raises_an_error_train_test():
+    smiles_list = ["CCC"]
+
+    with pytest.raises(
+        ValueError,
+        match="Train subset is empty.",
+    ):
+        _ = scaffold_train_test_split(smiles_list, use_csk=True)
```
j-adamczyk, line 218: You don't need `_ = ` here

j-adamczyk: Also, is `use_csk=True` needed?

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -144,16 +144,16 @@ def randomized_scaffold_train_test_split(
 
     train_idxs: list[int] = []
     test_idxs: list[int] = []
-    desired_test_size = int(test_size * len(data))
+    desired_test_size = max(1, int(test_size * len(data))) if test_size > 0 else 0
```
j-adamczyk, line 147: 1. If train or test size is 0, we should raise an error
2. I would prefer to modify `validate_train_test_sizes`. The basic problem is that we allow 2 types of inputs for `test_size`: int with absolute number of rows or float with proportion. If `validate_train_test_sizes` always returned the integer number of rows, you would not need this line here at all. I think it would quite simplify the code

j-adamczyk (changes requested):

skfp/model_selection/randomized_scaffold_split.py
StaniszewskiA: So, am I supposed to move this functionality into validate_train_test_sizes, which would result in it returning an integer?

j-adamczyk: Basically yes. I would also explicitly check if `test_size` is in range `(0, 1)` - this will tell you if it's a float and thus a proportion, rather than an integer (number of molecules).

StaniszewskiA: I believe that range check is already implemented inside the **validate_params** decorator.

j-adamczyk: `validate_params` checks if it's an integer OR a float in that range, that is correct. But it's just a type check. You have to have 2 separate behaviors there in `validate_train_test_sizes`:
- if it's integer, check the size (can't be larger than the dataset)
- if it's a float, turn into integer and return

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -369,13 +355,17 @@ def _create_scaffolds(
...
     for idx, mol in enumerate(molecules):
+        mol = deepcopy(mol)
+        Chem.RemoveStereochemistry(mol)  # important for canonization of CSK!
+        scaff = MurckoScaffold.GetScaffoldForMol(mol)
+        scaff = MurckoScaffold.MakeScaffoldGeneric(scaff)
+        scaff = MurckoScaffold.GetScaffoldForMol(scaff)
+
         if use_csk:
-            scaffold = MurckoScaffold.MakeScaffoldGeneric(mol=mol)
+            scaff = MurckoScaffold.MakeScaffoldGeneric(mol=mol)
```
j-adamczyk, line 365: `scaffold` was nice

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -369,13 +355,17 @@ def _create_scaffolds(
     molecules = ensure_mols(data)
 
     for idx, mol in enumerate(molecules):
+        mol = deepcopy(mol)
+        Chem.RemoveStereochemistry(mol)  # important for canonization of CSK!
+        scaff = MurckoScaffold.GetScaffoldForMol(mol)
+        scaff = MurckoScaffold.MakeScaffoldGeneric(scaff)
+        scaff = MurckoScaffold.GetScaffoldForMol(scaff)
```
j-adamczyk, line 362: Too much copy-paste, you always should do `RemoveSterochemistry`, rest is only for CSK

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -357,7 +343,7 @@ def randomized_scaffold_train_valid_test_split(
 
 
 def _create_scaffolds(
-    data: Sequence[Union[str, Mol]], include_chirality: bool = False, use_csk=False
+    data: Sequence[Union[str, Mol]], use_csk=False
```
j-adamczyk, line 346: `use_csk: bool = False`

skfp/model_selection/randomized_scaffold_split.py
```diff
@@ -131,8 +127,10 @@ def randomized_scaffold_train_test_split(
 
     """
     # flake8: noqa: E501
-    train_size, test_size = validate_train_test_sizes(train_size, test_size)
-    scaffolds = _create_scaffolds(data, include_chirality, use_csk)
+    train_size, test_size = validate_and_scale_train_test_sizes(
```
j-adamczyk, line 130: I think previous name was ok, no need for `_and_scale` (scaling has particular meaning in ML, after all)

skfp/model_selection/utils.py
```diff
@@ -1,41 +1,74 @@
...
-def validate_train_test_sizes(
-    train_size: Optional[float], test_size: Optional[float]
+def validate_and_scale_train_test_sizes(
+    train_size: Optional[Union[float, int]],
+    test_size: Optional[Union[float, int]],
+    data_length: int,
 ) -> tuple[float, float]:
     """
     Fill in missing sizes for train and test sets based on the provided sizes.
+    If test_size and train_size are floats, this method returns scaled data_length.
```
j-adamczyk, line 24: `this method returns concrete number of rows, i.e. int(size * data_length)`

skfp/model_selection/utils.py
```diff
@@ -1,41 +1,74 @@
...
+    If test_size and train_size are ints, this method returns the provided sizes.
     """
     if train_size is None and test_size is None:
-        return 0.8, 0.2
+        train_size, test_size = 0.8, 0.2
+
+    if (
+        train_size is not None
+        and test_size is not None
+        and type(train_size) is not type(test_size)
```
j-adamczyk, line 33: Is this covered by tests?

skfp/model_selection/utils.py
```diff
@@ -1,41 +1,74 @@
...
+        else:
+            raise ValueError("test_size must be provided when train_size is None.")
+    elif isinstance(test_size, int):
+        if test_size >= data_length:
+            raise ValueError(
+                "test_size as an integer must be smaller than data_length."
+            )
+
+    if isinstance(train_size, float) and isinstance(test_size, float):
+        if not abs(train_size + test_size - 1.0) < 1e-9:
```
j-adamczyk, line 60: `if not np.isclose(train_size + test_size, 1.0)`

skfp/model_selection/utils.py
```diff
@@ -1,41 +1,74 @@
...
+    if train_size == 0.0:
+        raise ValueError("train_size is 0.0")
+    if test_size == 0.0:
+        raise ValueError("test_size is 0.0")
 
-    return train_size, test_size
+    if isinstance(train_size, float):
+        return train_size * data_length, test_size * data_length
+    else:
+        return float(train_size), float(test_size)
```
j-adamczyk, line 71: Why convert `int` to `float`, if it was explicitly given?

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+        valid_size,
+        test_size,
+    )
 
-    if valid_size == 0.0:
-        warnings.warn(
-            "Validation set will not be returned since valid_size was set to 0.0."
-            "Consider using train_test_split instead."
+    if not all(isinstance(size, (int, float)) for size in sizes):
+        raise ValueError("All sizes must be either int or float.")
```
j-adamczyk, line 122: Print types if wrong:
```
types = [type(size) for size in sizes)
raise ... or float, got: {types}"
```

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+
+    if not all(isinstance(size, type(sizes[0])) for size in sizes):
+        raise ValueError("All sizes must be of the same type.")
```
j-adamczyk, line 125: as above

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+    )
 
-    if valid_size == 0.0:
-        warnings.warn(
-            "Validation set will not be returned since valid_size was set to 0.0."
-            "Consider using train_test_split instead."
+    if not all(isinstance(size, (int, float)) for size in sizes):
+        raise ValueError("All sizes must be either int or float.")
+
+    if not all(isinstance(size, type(sizes[0])) for size in sizes):
```
j-adamczyk, line 124: Replace `sizes[0]` with `train_size`

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+        raise ValueError("All sizes must be of the same type.")
+
+    if isinstance(sizes[0], float):
```
j-adamczyk, line 127: Replace `sizes[0]` with `train_size`

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+        if any(size < 0 for size in sizes):
+            raise ValueError("Sizes as floats must be non-negative.")
```
j-adamczyk, line 129: We assume it is so here, since this function is already called after a function with validator. You already make this assumption in train-test split

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+
+        total = sum(sizes)
+        if not (abs(total - 1.0) < 1e-9):
```
j-adamczyk, line 132: `if not np.isclose(sum(sizes), 1.0):`

skfp/model_selection/utils.py
```diff
@@ -63,29 +95,61 @@ def split_additional_data(
...
+            raise ValueError(
+                "The sum of train_size, valid_size, and test_size must be 1.0."
```
j-adamczyk, line 134: `, got: {sum(sizes)}`

skfp/model_selection/utils.py
```diff
@@ -1,41 +1,74 @@
...
+        if isinstance(test_size, float):
             train_size = 1 - test_size
         else:
             raise ValueError("test_size must be provided when train_size is None")
-    elif test_size is None:
-        test_size = 1 - train_size
+    elif isinstance(train_size, int):
+        if train_size >= data_length:
+            raise ValueError(
+                "train_size as an integer must be smaller than data_length."
```
j-adamczyk, line 45: `than data_length, got {train_size} for data {data_length}"`

j-adamczyk: And don't add dots at the end of errors

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -40,6 +45,24 @@ def additional_data() -> list[list[Union[str, int, bool]]]:
     ]
 
 
+@pytest.fixture
+def smiles_for_ten_different_scaffold() -> list[str]:
```
j-adamczyk, line 49: `smiles_ten_scaffolds` is enough

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -80,7 +103,7 @@ def test_csk_should_fail_for_degree_greater_than_four():
         Chem.rdchem.AtomValenceException,
         match="Explicit valence for atom # 1 C, 6, is greater than permitted",
     ):
-        _ = randomized_scaffold_train_test_split(smiles, random_state=42, use_csk=True)
+        randomized_scaffold_train_test_split(smiles, random_state=42, use_csk=True)
```
j-adamczyk, line 106: Prefer `random_state=0` everywhere for uniformity

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -150,3 +173,60 @@ def test_seed_consistency_train_valid__test_split(all_molecules):
...
+        data=smiles_for_ten_different_scaffold, train_size=7, test_size=3, use_csk=True
+    )
+    assert len(train_set) == 7
+    assert len(test_set) == 3
+
+
+def test_train_test_split_properly_splits_csk_with_floats(
+    smiles_for_ten_different_scaffold,
+):
+
```
j-adamczyk, line 192: Unnecessary empty line

j-adamczyk: Same below

tests/model_selection/scaffold_split.py
```diff
@@ -206,3 +224,85 @@ def test_scaffold_train_valid_test_split_return_indices(all_molecules):
     assert all(isinstance(idx, int) for idx in train_idxs)
     assert all(isinstance(idx, int) for idx in valid_idxs)
     assert all(isinstance(idx, int) for idx in test_idxs)
+
```
j-adamczyk, line 227: Also add tests for raising errors - we have quite a few more situations where errors can be raised, and they should be covered by tests. You can use test coverage (see Makefile) to check which lines are not tested.

j-adamczyk (changes requested):

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -34,11 +39,65 @@ def all_molecules() -> list[str]:
...
+            False,
+            True,
+            False,
+            True,
+            False,
+            True,
+            False,
+            True,
+        ],
+    ]
```
j-adamczyk, line 81: Does this have to be such a long list? If so, a shorter code:
```
import string

num = 18
ints = list(range(1, num))
letters = list(string.ascii_lowercase)[:num]
bools = [i % 2 == 0 for i in range(num)]
return [ints, letters, bools]
```

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -73,30 +132,16 @@ def test_randomized_scaffold_count_for_xanthines():
     assert len(randomized_scaffolds) == 1
 
 
-def test_csk_should_fail_for_degree_greater_than_four():
+def test_csk_should_not_fail_for_degree_greater_than_four():
```
j-adamczyk, line 135: Shorter: `test_csk_high_degree_atoms`

tests/model_selection/randomized_scaffold_split.py
```diff
@@ -140,13 +185,86 @@ def test_seed_consistency_train_test_split(all_molecules):
...
+
+    assert isinstance(train_idxs, list), "Train indices should be a list"
+    assert isinstance(test_idxs, list), "Test indices should be a list"
+
+    assert all(
+        isinstance(idx, int) for idx in train_idxs
+    ), "All train indices should be integers"
+    assert all(
+        isinstance(idx, int) for idx in test_idxs
+    ), "All test indices should be integers"
```
j-adamczyk, line 270: No need for string errors here - if something fails, this should be obvious from the code. And it is IMO, since the code is nice and readable

j-adamczyk (approved):

skfp/model_selection/utils.py
```diff
@@ -64,28 +99,64 @@ def split_additional_data(
...
+                f"The sum of train_size, valid_size, and test_size must be 1.0, got: "
+                f"{sum(sizes)}"
+            )
 
-    return train_size, valid_size, test_size
+        train_size_int = int(train_size * data_length)
+        valid_size_int = max(int(valid_size * data_length), 1)
+        test_size_int = max(int(data_length - train_size_int - valid_size_int), 1)
+
+        return (int(train_size_int), int(valid_size_int), int(test_size_int))
```
j-adamczyk, line 144: You don't need another `int` cast here, you do this directly above.

skfp/model_selection/utils.py
```diff
@@ -64,28 +99,64 @@ def split_additional_data(
...
+    if isinstance(train_size, float):
+        if not np.isclose(sum(sizes), 1.0):
+            raise ValueError(
+                f"The sum of train_size, valid_size, and test_size must be 1.0, got: "
+                f"{sum(sizes)}"
+            )
 
-    return train_size, valid_size, test_size
+        train_size_int = int(train_size * data_length)
+        valid_size_int = max(int(valid_size * data_length), 1)
```
j-adamczyk, line 141: Alternatively: `int(max(valid_size * data_length, 1))`

skfp/model_selection/utils.py
j-adamczyk, line 144: You don't need another `int` cast here, you do this directly above.

skfp/model_selection/utils.py
j-adamczyk, line 141: Alternatively: `int(max(valid_size * data_length, 1))`

skfp/model_selection/utils.py
```diff
@@ -64,28 +99,64 @@ def split_additional_data(
...
+
+        return int(train_size), int(valid_size), int(test_size)
+
+
+if __name__ == "__main__":
+    data = ["C1CCCC(C2CC2)CC1", "c1n[nH]cc1C1CCCCCC1"]
+    result = validate_train_valid_test_split_sizes(
+        None, None, None, data_length=len(data)
+    )
+    print(result)
```
j-adamczyk, line 162: Leftover from debugging?

j-adamczyk (changes requested):

j-adamczyk (approved):

PR #244, opened 2024-09-06
Add PR template and update CONTRIBUTING.md

j-adamczyk:


mjste (approved):

mjste:
Maybe add: check if docs render properly on localhost (when moving modules, changing do strings etc)

Issue #197, opened 2024-07-13
Add molecular filters


j-adamczyk:
Molecular filters are simple rule sets that assume that viable molecules have to conform to a some basic statistical properties, like molecular weight or number of rotatable bonds. They can be used as data quality checks, or initial filtering tools. There are quite a few of them available, both in RDKit and elsewhere, and they are very simple to implement. Some filters are also known as rules, rule sets, or structural alerts.

Example molecular filters:
- [x] Lipinski Rule of 5 - [ref](https://www.rdkit.org/docs/GettingStartedInPython.html#lipinski-rule-of-5)
- [x] PAINS (A, B, C) - [ref 1](https://www.rdkit.org/docs/GettingStartedInPython.html#filtering-unwanted-substructures), [ref 2](https://github.com/rdkit/rdkit/blob/e4f4644a89d6446ddebda0bf396fa4335324c41c/Code/GraphMol/FilterCatalog/Filters.cpp#L506), [ref 3](https://www.soci.org/-/media/files/conference-downloads/2011/designing-safer-medicines-in-discvery-mar-2011/francis_atkinson_presentation.ashx)
- [x] Brenk - [ref 1](https://www.rdkit.org/docs/GettingStartedInPython.html#filtering-unwanted-substructures), [ref 2](https://github.com/rdkit/rdkit/blob/e4f4644a89d6446ddebda0bf396fa4335324c41c/Code/GraphMol/FilterCatalog/Filters.cpp#L506), [ref 3](https://www.soci.org/-/media/files/conference-downloads/2011/designing-safer-medicines-in-discvery-mar-2011/francis_atkinson_presentation.ashx)
- [x] NIH - [ref 1](https://www.rdkit.org/docs/GettingStartedInPython.html#filtering-unwanted-substructures), [ref 2](https://github.com/rdkit/rdkit/blob/e4f4644a89d6446ddebda0bf396fa4335324c41c/Code/GraphMol/FilterCatalog/Filters.cpp#L506), [ref 3](https://www.soci.org/-/media/files/conference-downloads/2011/designing-safer-medicines-in-discvery-mar-2011/francis_atkinson_presentation.ashx)
[...]

PR #247, opened 2024-09-07
Lipinski filter

j-adamczyk:
## Changes

Added Lipinski Rule of 5 filter. With `BaseFilter` class, it can serve as a base for issue https://github.com/scikit-fingerprints/scikit-fingerprints/issues/197.

### DIFF
--- a/skfp/preprocessing/filters/lipinski.py
+++ b/skfp/preprocessing/filters/lipinski.py
@@ -1,0 +1,97 @@
+from typing import Optional
+
+from rdkit.Chem import Mol
+from rdkit.Chem.Crippen import MolLogP
+from rdkit.Chem.Descriptors import MolWt
+from rdkit.Chem.rdMolDescriptors import CalcNumLipinskiHBA, CalcNumLipinskiHBD
+
+from skfp.bases.base_filter import BaseFilter
+
+
+class LipinskiFilter(BaseFilter):
+    """
+    Lipinski's Rule of 5 (RO5).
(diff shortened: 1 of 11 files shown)


skfp/preprocessing/filters/lipinski.py
```diff
@@ -0,0 +1,97 @@
...
+    Examples
+    --------
+    >>> from skfp.preprocessing import LipinskiFilter, MolToSmilesTransformer
+    >>> smiles = ["[C-]#N", "CC=O", "O=C(O)c1ccccc1c2ccc(cc2)Cn3c4cc(cc(c4nc3CCC)C)c5nc6ccccc6n5C"]
+    >>> filt = LipinskiFilter()
+    >>> filt
+    LipinskiFilter()
+
+    >>> filtered_mols = filt.transform(smiles)
+    >>> mol_to_smiles = MolToSmilesTransformer()
```
mjste, line 64: There's no need for MolToSmilesTransformer -  filter returns SMILES anyway.

mjste (changes requested):
Add base_filter to docs

skfp/preprocessing/filters/lipinski.py
j-adamczyk: Right, I forgot about that

mjste (changes requested):
Add BaseFilter to docs

mjste (approved):

Issue #178, opened 2024-06-09
Add data splitting functions


j-adamczyk:
Currently, we have no data splitting functionalities. They are often tightly coupled with fingerprint computation, and essential for molecular property prediction and de novo drug design in general, so it would be a good addition. See [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC10575948/) for general reference.

Splitters should be regular functions, similar to `train_test_split` in scikit-learn. They should take lists (or sequences in general) of SMILES strings, or `Mol` objects, or computed fingerprints. If SMILES or `Mol` are passed, fingerprints should be calculated. Those functions must be kept simple, even if they rely on molecular fingerprints, since their parameters are not something to optimize. Reasonable defaults from papers or other libraries are important here.

Splitters to implement:
- [x] scaffold split - [ref 1](https://arxiv.org/abs/1703.00564), [ref 2](https://github.com/deepchem/deepchem/blob/ebb7824f4b99fa95f606a2a60c37a4edd0d52932/deepchem/splits/splitters.py#L1481)
- [x] randomized scaffold split - [ref 1](https://arxiv.org/abs/2207.06010); note the important difference from scaffold split, the randomized or "balanced" version is nondeterministic, and must support `random_state` argument
- [x] MaxMin split - [ref 1](https://rdkit.blogspot.com/2017/11/revisting-maxminpicker.html), [ref 2](https://arxiv.org/abs/2404.16196), [ref 3](https://onlinelibrary.wiley.com/doi/abs/10.1002/qsar.200290002), [ref 4](https://github.com/deepchem/deepchem/blob/ebb7824f4b99fa95f606a2a60c37a4edd0d52932/deepchem/splits/splitters.py#L1037)
[...]

PR #271, opened 2024-09-23
Implemented Butina split

j-adamczyk:
## Changes

Butina split, with quite a few optimization, including approximate version. Also contains fixes for numerical aspects of Electroshape fingerprint, since it errored during testing. Covers part of https://github.com/scikit-fingerprints/scikit-fingerprints/issues/178.


skfp/model_selection/splitters/butina_split.py
```diff
@@ -0,0 +1,448 @@
...
+    fingerprints with 2048 bits are used as features.
+    """
+    mols = ensure_mols(data)
+
+    fps_rdkit = GetMorganGenerator().GetFingerprints(mols)
+    centroid_idxs = LeaderPicker().LazyBitVectorPick(
+        fps_rdkit, poolSize=len(mols), threshold=threshold
+    )
+    centroid_idxs = list(centroid_idxs)
+    non_centroid_idxs = sorted(set(range(len(mols))) - set(centroid_idxs))
```
my-alaska, line 403: GetFingerprints returns a tuple of `ExplicitBitVect`. Wouldn't it be easier to convert the results yielded by the skfp transformer to `ExplicitBitVect` instead of converting the datasets with Morgan generator?

j-adamczyk: No, `ExplicitBitVect` is a C++ object and cannot be created from Python. This is the only way we can do this, as far as I know

mjste (approved):

PR #284, opened 2024-10-06
Time benchmark improvements

my-alaska:
## Changes

Small improvements and changes in time benchmark.
Added:
- Stabilizing measurements with garbage collector
- Saving to pdf
- Vertical line at value 1 for speedup plot

- [ ] Docstrings added/updated in public functions and classes (N/A)
- [ ] Tests added, reasonable test coverage (at least ~90%, `make test-coverage`) (N/A)
- [ ] Sphinx docs added/updated and render properly (`make docs` and see `docs/_build/index.html`) (N/A)


j-adamczyk (approved):

mjste (changes requested):
1. Would it be possible to remove ogb? We already have datasets in skfp, and ogb is a massive dependency (depends on torch)
2. Add needed dependencies. We have barely needed ogb, but not matplotlib which is essential (group eval)

Issue #310, opened 2024-10-17
Add descriptor names to fingerprints


j-adamczyk:
We should support descriptor names for fingerprints where possible. This was also mentioned in https://github.com/scikit-fingerprints/scikit-fingerprints/issues/307. Overriding `get_feature_names_out` in inheriting classes should be enough to make this work.

In general:
1. Hashed fingerprints (e.g. ECFP) and other fingerprints with variable number of features - leave as is, they don't have interpretable features
2. Substructure with SMARTS only (e.g. Klekota-Roth) - use SMARTS as feature names
3. Substructure with interpretable names (e.g. Laggner) - use those names
4. Descriptor-based (e.g. Mordred, autocorrelation, functional groups) - use appropriate feature names

List of fingerprints to cover:

- [x] AutocorrFingerprint
- [x] ElectroShapeFingerprint
- [x] EStateFingerprint
- [x] FunctionalGroupsFingerprint
- [x] GETAWAYFingerprint
- [x] GhoseCrippenFingerprint
- [x] KlekotaRothFingerprint
- [x] LaggnerFingerprint
- [x] MACCSFingerprint
- [x] MordredFingerprint
- [x] MORSEFingerprint
- [x] MQNsFingerprint
- [x] PubChemFingerprint
- [x] RDFFingerprint
- [x] RDKit2DDescriptorsFingerprint
- [x] USRFingerprint
- [x] USRCATFingerprint
- [x] VSAFingerprint
- [x] WHIMFingerprint

j-adamczyk:
@Hrovatin on one hand this makes sense, on the other there are tricky cases. For example, pharmacophore uses histograms of distances between different pharmacophores. It would be very hard to interpret - is this meaningful? Similar case for MORSE and Autocorr, for example, which use convoluted feature definitions with histogram bins. Klekota-Roth fingerprint uses enumerated SMARTS (auto-generated in their original paper) with no actual interpretation, whereas e.g. Laggner fingerprint has hand-made SMARTS with human-readable interpretation.

I am not against the idea of somehow classifying fingerprints with additional attributes, possibly with feature readability, but I think there are many edge cases here, and we give this some thought to do this properly.

PR #312, opened 2024-10-20
Pass Through Descriptor Names for Mordred / RDKit2DDescriptors

Scienfitz:
## Changes

Progress for #310 
Pass through descriptor names from underlying mordred / rdkit2ddescriptor calculators via `get_feature_names_out`

## Notes
- [base class implementation](https://github.com/scikit-learn/scikit-learn/blob/a072e56fa/sklearn/base.py#L898) for reference
- [Mordred treatment of descriptor names](https://github.com/JacksonBurns/mordred-community/blob/23a3a3967968e9585b821de97ffca6095ff580d7/mordred/_base/calculator.py#L173)
- [RDKit treatment of descriptor names](https://github.com/bp-kelley/descriptastorus/blob/master/descriptastorus/descriptors/rdDescriptors.py#L439)
- Coverage: 
image

### DIFF
--- a/skfp/fingerprints/mordred_fp.py
+++ b/skfp/fingerprints/mordred_fp.py
@@ -123,7 +123,8 @@ class MordredFingerprint(BaseFingerprintTransformer):
         )
 
     def get_feature_names_out(self):
-        """Get output feature names for transformation.
+        """
+        Get fingerprint output feature names.
 
         Returns
         -------
--- a/skfp/fingerprints/mordred_fp.py
+++ b/skfp/fingerprints/mordred_fp.py
@@ -122,27 +122,14 @@ class MordredFingerprint(BaseFingerprintTransformer):
             else np.array(X, dtype=np.float32)
         )
 
-    def get_feature_names_out(self, input_features=None):
+    def get_feature_names_out(self):
         """Get output feature names for transformation.
-
-        Parameters
-        ----------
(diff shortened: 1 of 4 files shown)


j-adamczyk (changes requested):
Code itself looks good, tests also. But please change docstrings - they are copy-paste from scikit-learn, and some things are done differently here. In particular, you can describe what names are used as feature names.

Scienfitz:
@j-adamczyk I've removed the call to `super` and the argument which had no purpose in the overwritten methods was kicked
This has also simplified docstrings.

Regarding a description what the feature names are: this is not going to be possible as these descriptors are large heterogeneous collections. So not sure what to do about this

j-adamczyk:
@Scienfitz sure, removing the argument makes sense. Regarding descriptions, we don't need explicit descriptions of all features, this should also be evident from the general fingerprint documentation. However, I think it can be slightly more specific.

Currently, it is `"Get output feature names for transformation."`. It could be e.g. `"Get fingerprint feature names, i.e. names of descriptors at given columns."`. This description should work for all descriptor fingerprints (e.g. Mordred, RDKit2D).

j-adamczyk (approved):

j-adamczyk:
@Scienfitz I made two small changes, everything looks nice, tests pass, so I will merge this. Thanks for the contribution!

PR #318, opened 2024-10-28
PubChem split

JPoziemski:
## Changes


skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    n_jobs : int, default=5
+        The number of jobs to run in parallel for fetching data from PubChem. Note that
+        since those are requests relying on I/O, threads are used instead of processes.
+        At most 5 requests per second are made to avoid throttling, which limits
+        parallelism even if high ``n_jobs`` is set.
+
+    Returns
+    ----------
+    subsets : tuple[list, list, ...]
+    Tuple with train-test subsets of provided arrays. First two are lists of SMILES
```
j-adamczyk, line 114: This should have 1 tab more

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    data_df = data_df[~data_df["year"].isna()]
+
+    data_df = data_df.sort_values("year")
+
+    train_size, test_size = validate_train_test_split_sizes(
+        train_size, test_size, len(data_df)
+    )
+
+    train_idxs: list[int] = list(data_df.iloc[:train_size].idx)
+    test_idxs: list[int] = list(data_df.iloc[train_size:].idx)
```
j-adamczyk, line 140: Is this the same as `.index`? If so, I find that more readable

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+        Nucleic Acids Res. 2018 Jul 2;46(W1):W563-W570.
+        <https://doi.org/10.1093/nar/gky294>`_
+    """
+    years = _get_pubchem_years(data, n_jobs)
+    data_df = pd.DataFrame({"idx": list(range(len(data))), "year": years})
+
+    if not_found_behavior != "remove":
+        not_found_mols = list(data_df[data_df["year"].isna()].idx)
+
+    data_df = data_df[~data_df["year"].isna()]
```
j-adamczyk, line 131: This is controlled by `not_found_behavior`. This line should only happen for `remove`. Note that `.sort_values` takes `na_position` argument, which will be useful for `train` and `test` values.

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    )
+
+    train_idxs: list[int] = list(data_df.iloc[:train_size].idx)
+    test_idxs: list[int] = list(data_df.iloc[train_size:].idx)
+
+    if not_found_behavior == "train":
+        train_idxs.extend(not_found_mols)
+
+    if not_found_behavior == "test":
+        test_idxs.extend(not_found_mols)
```
j-adamczyk, line 146: This can be if/elif/else (train/test/remove)

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    n_jobs : int, default=5
+        The number of jobs to run in parallel for fetching data from PubChem. Note that
+        since those are requests relying on I/O, threads are used instead of processes.
+        At most 5 requests per second are made to avoid throttling, which limits
+        parallelism even if high ``n_jobs`` is set.
+
+    Returns
+    ----------
+    subsets : tuple[list, list, ...]
+    Tuple with train-valid-test subsets of provided arrays. First three are lists of
```
j-adamczyk, line 277: Tab

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    data = list(
+        map(
+            lambda mol: (
+                MolToSmiles(mol, canonical=True, isomericSmiles=True)
+                if mol is Mol
+                else mol
+            ),
+            data,
+        )
+    )
```
j-adamczyk, line 355: Use `ensure_smiles`. Also, aren't `canonical=True, isomericSmiles=True` defaults?

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+
+    print("Converting SMILES to CIDs")
+    cids = run_in_parallel(get_cid_for_smiles, data, n_jobs)
+
+    print("Searching for CIDs literature")
```
j-adamczyk, line 360: Printing should be controlled with `verbose`

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    years = run_in_parallel(get_earliest_publication_date, cids, n_jobs)
+
+    return years
+
+
+def get_cid_for_smiles(smiles: str) -> Optional[str]:
```
j-adamczyk, line 366: I think this can have `_` at the beginning, it's pretty internal function. And then docstring can be a minimal comment, we don't need whole `Parameters` and `Return` format for internal functions.

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    ------------
+    smiles: str
+        SMILES string
+
+    Return
+    --------
+        PubChem compound identifier if  else None
+
+    """
+    PUG_url = f"https://pubchem.ncbi.nlm.nih.gov/rest/pug/compound/smiles/{quote(smiles)}/cids/JSON"
```
j-adamczyk, line 379: You don't need `PUG` prefixes here, they are obvious from context

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+
+    PUG_response = requests.get(PUG_url, timeout=10)
+    PUG_response_status = PUG_response.status_code
+
+    if PUG_response_status != 200:
+        warnings.warn(
+            f"PUG error \n, {PUG_url} \n status code {PUG_response_status}, \n check your input"
+        )
```
j-adamczyk, line 387: PUG can easily return errors on server overloaded error. This has to be resilient and implement retries. They may also be configurable in the splitter, e.g. 10 retries by default, with 1 second wait in between.

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    PUG_response_json = PUG_response.json()
+
+    if "IdentifierList" in PUG_response_json:
+        cid = PUG_response_json["IdentifierList"]["CID"][0]
+        if bool(cid):
+            cid = str(cid)
+        else:
+            cid = None
+    else:
+        cid = None
```
j-adamczyk, line 398: Shorter:
```
try:
    cid = PUG_response_json["IdentifierList"]["CID"][0]
    cid = str(cid) if cid else None
except KeyError, ValueError:
    cid = None
```

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    return cid
+
+
+def get_earliest_publication_date(cid: Union[str, None]) -> Optional[int]:
```
j-adamczyk, line 402: `cid: Optional[str]`

j-adamczyk: Function can be private

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    def get_publication_date(response: list[dict]) -> Optional[int]:
+        article_pub_date = None
+        for entry in response:
+            if "articlepubdate" in entry:
+                return int(entry["articlepubdate"][:4])  # get year of publication
+            else:
+                continue
+        return article_pub_date
+
+    while True:
```
j-adamczyk, line 441: This should be limited, like the function above, with configurable num retries and timeout

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+    params = {"infmt": "json", "outfmt": "json", "query": json.dumps(query)}
+
+    def get_publication_date(response: list[dict]) -> Optional[int]:
+        article_pub_date = None
+        for entry in response:
+            if "articlepubdate" in entry:
+                return int(entry["articlepubdate"][:4])  # get year of publication
+            else:
+                continue
+        return article_pub_date
```
j-adamczyk, line 439: Can be simpler:
```
        for entry in response:
            if "articlepubdate" in entry:
                return int(entry["articlepubdate"][:4])  # get year of publication
```

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+            if isinstance(response, list):
+                if len(response) >= 1:
+                    date = get_publication_date(response)
+                    return date
+                else:
+                    # no literature found
+                    return None
+            else:
+                # most probably "Server too busy" error
+                time.sleep(1)
```
j-adamczyk, line 453: We should probably check for this explicitly

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,456 @@
...
+
+    return_indices : bool, default=False
+        Whether the method should return the input object subsets, i.e. SMILES strings
+        or RDKit ``Mol`` objects, or only the indices of the subsets instead of the data.
+
+    return_indices : bool, default=False
+        Whether the method should return the input object subsets, i.e. SMILES strings
+        or RDKit `Mol` objects, or only the indices of the subsets instead of the data.
+
+    n_jobs : int, default=5
```
j-adamczyk, line 268: You should check this somewhere and explicitly limit this, so that we don't exceed 5 requests per second

skfp/model_selection/splitters/utils.py
```diff
@@ -149,3 +151,30 @@ def validate_train_valid_test_split_sizes(
...
+        super().__init__(*args, **kwargs)
+        self.total = total
+
+    def __call__(self, *args, **kwargs):
+        with tqdm(total=self.total) as self._pbar:
+            return Parallel.__call__(self, *args, **kwargs)
+
+    def print_progress(self):
+        self._pbar.n = self.n_completed_tasks
+        self._pbar.refresh()
```
j-adamczyk, line 180: We already have those implemented, see `utils` module

tests/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+@pytest.fixture
+def proper_pubchem_cid() -> str:
+    return "24261"
+
+
+@pytest.fixture
+def get_none() -> None:
+    return None
```
j-adamczyk, line 38: Don't use fixture for this, just explicitly use `None`

tests/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+@pytest.fixture
+def smiles_that_not_exists_in_pubchem() -> str:
+    return "C[Fe](C)(C)OC(C)=O"
+
+
+@pytest.fixture
+def proper_pubchem_cid() -> str:
+    return "24261"
```
j-adamczyk, line 33: If you have fixtures that are used only in a single place, then don't use a fixture, but rather provide the value explicitly. This brings it closer to actual usage, increasing readability

j-adamczyk (changes requested):

skfp/model_selection/splitters/pubchem_split.py
JPoziemski: pubchem has restriction 5 requests per second  (https://pubchem.ncbi.nlm.nih.gov/docs/pug-rest)

mjste (approved):

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,447 @@
...
+        "nullatbottom": 1,
+    }
+    base_url = "https://pubchem.ncbi.nlm.nih.gov/sdq/sdqagent.cgi"
+    params = {"infmt": "json", "outfmt": "json", "query": json.dumps(query)}
+
+    def get_publication_date(response: list[dict]) -> Optional[int]:
+        article_pub_date = None
+        for entry in response:
+            if "articlepubdate" in entry:
+                return int(entry["articlepubdate"][:4])  # get year of publication
```
j-adamczyk, line 425: We should probably have try/except here, in case something wrong happens with the year on PubChem's side

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,447 @@
...
+        return article_pub_date
+
+    trial = 0
+    while trial < n_trials:
+        try:
+            response = requests.get(base_url, params=params, timeout=10).json()
```
j-adamczyk, line 431: What happens when timeout is reached? Don't we get some other exception? This should be checked

j-adamczyk (changes requested):

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,447 @@
...
+        SMILES string
+
+    Return
+    --------
+        PubChem compound identifier if  else None
+
+    """
+    url = f"https://pubchem.ncbi.nlm.nih.gov/rest/pug/compound/smiles/{quote(smiles)}/cids/JSON"
+
+    response = requests.get(url, timeout=10)
```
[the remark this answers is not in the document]
JPoziemski: 1. Retried should be controled by some parameter or hardcoded ? 
2. timeout 20 is ok ?
3. i think the reasoable way is to create loop in range(1, retries) and try excepted on timeout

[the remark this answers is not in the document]
j-adamczyk: 1. Number of retries should be controllable
2. Yes, for a single request timeout 20 is ok
3. Makes sense

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,447 @@
...
+
+    def get_publication_date(response: list[dict]) -> Optional[int]:
+        article_pub_date = None
+        for entry in response:
+            if "articlepubdate" in entry:
+                return int(entry["articlepubdate"][:4])  # get year of publication
+        return article_pub_date
+
+    trial = 0
+    while trial < n_trials:
```
j-adamczyk, line 429: `for` loop would also work, right?

skfp/model_selection/splitters/pubchem_split.py
```diff
@@ -0,0 +1,447 @@
...
+                    return None
+            else:
+                # most probably "Server too busy" error
+                time.sleep(1)
+        except requests.ConnectionError:
+            # most probably unstable PubChem network
+            time.sleep(1)
+        trial += 1
+
+    return None
```
j-adamczyk, line 447: If all trials failed, I think we should raise an error, because it's not `None`, but rather a failure because of PubChem servers

j-adamczyk (changes requested):

Issue #303, opened 2024-10-13
MAP4 implementation is wrong


LucaCappelletti94:
Hi!

By chance, I believe I have found definitive proof that the [current implementation of MAP4 available in this library](https://github.com/scikit-fingerprints/scikit-fingerprints/blob/05326d0504ac9c18cc7e5924404ea12b2e280016/skfp/fingerprints/map.py#L21-L141) must contain errors.

My claim is based on two pieces of evidence: visualizations and model performance. Since the [original code](https://github.com/reymond-group/map4) is admittedly hard to use, [I have fully refactored it myself here](https://github.com/LucaCappelletti94/map4) and [published it on pip](https://pypi.org/project/map4/). I believe my implementation to pedissequally follow their original code and paper, minus some dumb points in their code (they were converting things to a set then to a list then to a set and other such things). I [have opened a pull request in their original repo](https://github.com/reymond-group/map4/pull/27), but they do not seem interested in replying.

By having a different implementation, I was able to compare your implementation with mine.

Both visualizations and performance report to the tasks I am working on, namely the prediction of chemical taxonomy given as input for the fingerprint. The names of the chemical taxonomy layers are "pathways", "superclasses" and "classes". Since chemicals may have multiple labels, the task is defined as a multi-label task.

## First evidence: Feature Visualizations

Here are the PCA and TSNE decompositions of the MAP implementation available from scikit-fingerprints, and [map4](https://github.com/LucaCappelletti94/map4), respectively. The colors are not some model predictions, but the ground truth labelling associated with this dataset ([dataset from this paper](https://pubs.acs.org/doi/10.1021/acs.jnatprod.1c00399)). While there is admittedly some uncertainty on the quality of some of these labels, I expect most of the labels to be correct. For each chemical taxonomical layer, the most common label is shown.

![Scikit-fingerprint's variant](https://github.com/LucaCappelletti94/hammer/blob/main/data_visualizations/MinHashed%20Atom%20Pair%20(2r,%202048b).png?raw=true)

![MAP4](https://github.com/LucaCappelletti94/hammer/blob/main/data_visualizations/MAP4.png?raw=true)

You can find several other such visualizations for other fingerprints made available from scikit-fingerprint [here](https://github.com/LucaCappelletti94/hammer/tree/main/data_visualizations). They all work nominally, with the singular exception of `MAP` which shows what I can describe by my experience as a near-total decomposition failure, both from the PCA and TSNE perspective.

[...]

PR #334, opened 2024-11-30
Added fixed version of map4

LucaCappelletti94:
## Changes

Resolved issue #303. Here follows the TSNEs of the older and newer versions - the improvement in quality is quite clear.

old_map
new_map

    - [x] Updated tests
    - [ ] Run coverage
- [ ] Sphinx docs added/updated and rendered properly (`make docs` and see `docs/_build/index.html`)

I am still unable to run coverage or any other such part. After another day with a colleague trying to make it work, I focused on doing the PR. Hopefully, at some point, I'll manage to set it up correctly.


skfp/fingerprints/map.py
```diff
@@ -139,135 +135,115 @@ def __init__(
         )
         self.fp_size = fp_size
         self.radius = radius
-        self.variant = variant
+        self.include_duplicated_shingles = include_duplicated_shingles
+        self._encoder: MHFPEncoder = MHFPEncoder(self.fp_size, seed=random_state)
```
j-adamczyk, line 139: This probably should be in `._calculate_fingerprint`. Note that if you e.g. do `.set_params()` with current code, the `MHFPEncoder` won't be updated, so this breaks scikit-learn compatibility.

j-adamczyk (changes requested):
Good PR in general, few changes requested.

In general, I think that we should not use `mhfp` library, but instead take the necessary parts of `MHFPEncoder` (https://github.com/reymond-group/mhfp/blob/master/mhfp/encoder.py) and create such class here, at the bottom of `map.py` file. We need only a few methods, and it can also probably be optimized.

Also, we should support the `count` variant. It's quite trivial, only requires addition instead of setting in the `.fold()` method of `MHFPEncoder`: https://github.com/reymond-group/mhfp/blob/master/mhfp/encoder.py#L343. Note that this requires `int` type, rather than `np.uint8` in many places, this is why we use it currently.

LucaCappelletti94:
> Good PR in general, few changes requested.
> 
> In general, I think that we should not use `mhfp` library, but instead take the necessary parts of `MHFPEncoder` (https://github.com/reymond-group/mhfp/blob/master/mhfp/encoder.py) and create such class here, at the bottom of `map.py` file. We need only a few methods, and it can also probably be optimized.

Yeah, arguably that org does not have the greatest track record in maintaining software considering that this refactoring was primarily started by me discovering their software didn't work. Better fold that in.

> Also, we should support the `count` variant. It's quite trivial, only requires addition instead of setting in the `.fold()` method of `MHFPEncoder`: https://github.com/reymond-group/mhfp/blob/master/mhfp/encoder.py#L343. Note that this requires `int` type, rather than `np.uint8` in many places, this is why we use it currently.

Ok, I'll try.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ rdkit = "<=2024.3.3"
 scikit-learn = "*"
 scipy = "*"
 tqdm = "*"
+mhfp = "*"
```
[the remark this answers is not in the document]
LucaCappelletti94: Removed entirely now.

skfp/fingerprints/map.py
```diff
@@ -139,135 +135,115 @@ def __init__(
...
-            idx = atom.GetIdx()
-            atom_envs = [
-                self._find_neighborhood(mol, idx, r) for r in range(1, self.radius + 1)
-            ]
-            atoms_env[idx].extend(atom_envs)
-
+            atom_identifier: int = atom.GetIdx()
+            for radius in range(1, self.radius + 1):
+                if atom_identifier not in atoms_env:
+                    atoms_env[atom_identifier] = []
```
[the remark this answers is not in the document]
LucaCappelletti94: Done

skfp/fingerprints/map.py
LucaCappelletti94: Oke

skfp/fingerprints/map.py
```diff
@@ -139,135 +144,124 @@ def __init__(
...
+            for shingling in shinglings:
+                hashed = struct.unpack("<I", sha256(shingling).digest()[:4])[0]
+                folded[hashed % self.fp_size] += 1
+            return folded
+
+        folded = np.zeros(self.fp_size, dtype=np.uint8)
+        for shingling in shinglings:
+            hashed = struct.unpack("<I", sha256(shingling).digest()[:4])[0]
+            folded[hashed % self.fp_size] = 1
+        return folded
```
j-adamczyk, line 176: I think that this can be deduplicated a bit:
```
dtype = np.uint32 if self.count else np.uint8
folded = np.zeros(self.fp_size, dtype=np.dtype)
for shingling in shinglings:
    hashed = struct.unpack("<I", sha256(shingling).digest()[:4])[0]
    idx = hashed % self.fp_size
    if self.count:
        folded[idx] += 1
    else:
        folded[idx] = 1
return folded
```

skfp/fingerprints/map.py
```diff
@@ -139,135 +144,124 @@ def __init__(
...
+        atom_identifier : int
+            The index of the atom to calculate the environment for
+        radius : int
+            The radius of the environment
+
+        Returns
+        -------
+        str
+            The SMILES representation of the atom environment, or
+            None if the atom is not found
```
j-adamczyk, line 195: We generally don't put full docstrings in private methods, since they would have to be maintained, and private methods in general could be changed without impacting the public API. Just the first sentence of this would suffice.

skfp/fingerprints/map.py
```diff
@@ -139,135 +144,124 @@ def __init__(
...
-            fp = np.bincount(bits, minlength=self.fp_size)
+        sub_molecule: Mol = PathToSubmol(
+            mol, atom_identifiers_within_radius, atomMap=atom_map
+        )
+        if atom_identifier in atom_map:
+            smiles = MolToSmiles(
+                sub_molecule,
+                rootedAtAtom=atom_map[atom_identifier],
+                canonical=True,
+                isomericSmiles=False,
```
j-adamczyk, line 210: Aren't those default values?

skfp/fingerprints/map.py
```diff
@@ -139,135 +144,124 @@ def __init__(
...
+            )
+            return smiles
 
-        return fp
+        return None
```
j-adamczyk, line 214: If you don't return anything, `None` will be returned anyway

skfp/fingerprints/map.py
```diff
@@ -48,10 +49,12 @@ class MAPFingerprint(BaseFingerprintTransformer):
         Number of iterations performed, i.e. maximum radius of resulting subgraphs.
         Another common notation uses diameter, therefore MAP4 has radius 2.
 
-    variant : {"raw_hashes", "bit", "count"}, default="bit"
-        Which variant to fingerprint to use. ``"raw_hashes"`` returns MinHash values.
-        ``"bit"`` folds values into binary vector, and ``"count"`` uses folding with
-        counting.
+    include_duplicated_shingles : bool, default=False
+        Whether to include duplicated shingles in the final fingerprint.
```
j-adamczyk, line 53: With folding and count variant, is this necessary? I mean, doesn't `count` argument already cover this?

j-adamczyk (changes requested):

skfp/fingerprints/map.py
LucaCappelletti94: The two operations are distinct: in the `count` case, you count all times a given bit in the bitvec is extracted by the hash. In the `include_duplicated_shingles`, you count as distinct features that you encounter the same shingle several times.

skfp/fingerprints/map.py
```diff
@@ -48,10 +49,12 @@ class MAPFingerprint(BaseFingerprintTransformer):
...
+
+    counts : bool, default=False
+        Whether to return counts of each shingle in the fingerprint, instead of
+        binary presence/absence.
```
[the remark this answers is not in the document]
LucaCappelletti94: Done

skfp/fingerprints/map.py
LucaCappelletti94: Kept only first line.

skfp/fingerprints/map.py
LucaCappelletti94: Refactored as proposed

skfp/fingerprints/map.py
LucaCappelletti94: Isn't it clearer if I leave it specified?

j-adamczyk: On one hand - absolutely, it is. On the other, if you provide values, at least I immediately think "oh, so they are different than defaults", since we typically don't provide those, at least for common functions. So I would prefer to just use defaults

skfp/fingerprints/map.py
j-adamczyk: So for the bit version there should be no difference, right? Since duplicate shingles should get hashed into the same index anyway. Count variant will have the difference, since:
- for `include_duplicated_shingles=False`, we only count hashing collisions between different subgraphs
- for `include_duplicated_shingles=True`, we will also include the same subgraph occurring multiple times

Am I right? If that's the case, I think we should just always include duplicates for `count`, and ignore this for `bit`. It just sounds very weird for me that someone would like to count bit collisions, but not same thing being present many times.

skfp/fingerprints/map.py
LucaCappelletti94: Turns out that `isomericSmiles` by default is True, so I left the other kwarg. It was like this in the original map4 impl, I am not certain why it was set this way.

skfp/fingerprints/map.py
```diff
@@ -48,10 +47,11 @@ class MAPFingerprint(BaseFingerprintTransformer):
...
         Another common notation uses diameter, therefore MAP4 has radius 2.
 
-    variant : {"raw_hashes", "bit", "count"}, default="bit"
-        Which variant to fingerprint to use. ``"raw_hashes"`` returns MinHash values.
-        ``"bit"`` folds values into binary vector, and ``"count"`` uses folding with
-        counting.
+    include_duplicated_shingles : bool, default=False
+        Whether to include duplicated shingles in the final fingerprint.
+
+    counts : bool, default=False
```
j-adamczyk, line 53: `count`, not `counts`, to follow other fingerprints

skfp/fingerprints/map.py
LucaCappelletti94: No, the `include_duplicated_shingles` makes it so that if you encounter X for the nth time you convert it into "X|nth", which makes all of them distinct.

LucaCappelletti94: This is as it was defined in the original impl btw, I am by no means saying it is something to keep.

skfp/fingerprints/map.py
j-adamczyk: Sure, just please leave a comment about this

skfp/fingerprints/map.py
LucaCappelletti94: Done

skfp/fingerprints/map.py
j-adamczyk: Ok, so your implementation is definitely correct, then. But I'm not sure if this is something really useful? Unless you see some immediate application, I would just remove this and do as I outlined above for simplicity. We don't have to follow the original implementation 100%, after all.

skfp/fingerprints/map.py
LucaCappelletti94: Done

skfp/fingerprints/map.py
LucaCappelletti94: It can conceivably have a better representation to distinguish some molecules that are characterized by strongly repeating patterns, it's something likely worth testing. That being said, I don't have a use case at hand where I could easily test this.

j-adamczyk: Fair point, I'll test it on some polymers. So leave it for now then, and I'm approving the PR

j-adamczyk (approved):

Issue #66, opened 2023-12-14
Add additional fingerprints


j-adamczyk:
We need to implement additional fingerprints from multiple sources, as we want scikit-fingerprints to be the only required Python library for computing molecular fingerprints.

RDKit:
- [ ] E-state - [ref](https://www.rdkit.org/docs/source/rdkit.Chem.EState.html)
- [ ] FPCP - [ref](https://www.rdkit.org/docs/GettingStartedInPython.html#morgan-fingerprints-circular-fingerprints)
- [ ] MHFP - potentially, benchmark with already implemented one, [ref](https://www.rdkit.org/docs/source/rdkit.Chem.rdMHFPFingerprint.html)
- [ ] Pattern - [ref](https://www.rdkit.org/docs/source/rdkit.Chem.rdmolops.html#rdkit.Chem.rdmolops.PatternFingerprint)
- [ ] Pharmacophore - [ref](https://www.rdkit.org/docs/GettingStartedInPython.html#d-pharmacophore-fingerprints)
- [ ] RDKit (substructure) - [ref](https://www.rdkit.org/docs/source/rdkit.Chem.rdFingerprintGenerator.html#rdkit.Chem.rdFingerprintGenerator.FPType.RDKitFP)

Descriptor-based:
- [ ] Atom Signature - [ref](https://cdk.github.io/cdk/latest/docs/api/org/openscience/cdk/fingerprint/SignatureFingerprinter.html)
- [ ] Hologram - [ref 1](https://jcheminf.biomedcentral.com/articles/10.1186/s13321-018-0316-5), [ref2](https://github.com/cdk/cdk/blob/main/descriptor/qsarmolecular/src/main/java/org/openscience/cdk/qsar/descriptors/molecular/JPlogPDescriptor.java)
- [ ] Klekota-Roth - [ref](https://cdk.github.io/cdk/latest/docs/api/org/openscience/cdk/fingerprint/KlekotaRothFingerprinter.html)
- [ ] Lingo - [ref](https://cdk.github.io/cdk/latest/docs/api/org/openscience/cdk/fingerprint/LingoFingerprinter.html)
- [ ] PubChem - [ref](https://deepchem.readthedocs.io/en/latest/api_reference/featurizers.html#pubchemfingerprint)

Other:
- [ ] Spectrophore - [ref](https://github.com/silicos-it/spectrophore)

my-alaska:
[...]

PR #371, opened 2025-01-25
Add BCUT2D fingerprint

j-adamczyk:
## Changes

BCUT2D is the last missing descriptor fingerprint from RDKit in https://github.com/scikit-fingerprints/scikit-fingerprints/issues/66. I rewrote this from scratch, since our version is now more flexible and allows other charge models than Gasteiger, which is hardcoded in RDKit and often fails. The downside is that it's about x2 slower than RDKit, but fortunately it's an extremely fast fingerprint anyway.


my-alaska (dismissed):

skfp/descriptors/charge.py
```diff
@@ -0,0 +1,71 @@
...
+    charges_problem = any(charge is None or np.isnan(charge) for charge in charges)
+    if charge_errors == "raise" and charges_problem:
+        smiles = MolToSmiles(mol)
+        raise ValueError(
+            f"Failed to compute at least one atom partial charge for {smiles}"
+        )
+    elif charge_errors == "zero":
+        charges = np.nan_to_num(np.array(charges, dtype=float), nan=0)
+    else:  # "ignore"
+        charges = [c for c in charges if c is not None]
```
mjste, line 69: Why don't we leave it as it is? I imagine a situation where the order and size of the charges may be actually meaningful (e.g. we know which atom failed.

my-alaska (approved):

PR #394, opened 2025-02-06
Add Simpson distance

StaniszewskiA:
## Changes


StaniszewskiA:
I've encountered some issues with Sphinx while trying to render the docs (hence the draft), but the code is ready for review.

skfp/distances/simpson.py
```diff
@@ -0,0 +1,193 @@
...
+    Computes the Simpson distance for binary data between two input arrays
+    or sparse matrices by subtracting the similarity from 1, using to
+    the formula:
+
+    .. math::
+        dist(vec_a, vec_b) = 1 - sim(vec_a, vec_b)
+
+    The calculated distance falls within the range ``[0, 1]``.
+    Passing all-zero vectors to this function results in a distance of 0.
+
```
j-adamczyk, line 126: We also include literature references in distance functions. Please add them here, and also for Rand similarity. Similarity and distance functions generally should have the same literature references.

skfp/distances/simpson.py
```diff
@@ -0,0 +1,193 @@
...
+    >>> vec_a = csr_array([[1, 0, 1]])
+    >>> vec_b = csr_array([[1, 0, 1]])
+    >>> dist = simpson_binary_distance(vec_a, vec_b)
+    >>> dist  # doctest: +SKIP
+    0.0
+    """
+    return 1 - simpson_binary_similarity(vec_a, vec_b)
+
+
+@njit(parallel=True)
```
j-adamczyk, line 161: Please measure if this results in real improvement on some datasets, e.g. BACE or HIV. Numba JIT has some initial overhead, and pure NumPy function may not benefit from it particularly.

tests/distances/simpson.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.mark.parametrize(
+    "vec_a, vec_b",
+    [
+        (np.array([1, 0, 0, 0]), np.array([1, 1, 1, 1])),
+        (csr_array([[1, 0, 0, 0]]), csr_array([[1, 1, 1, 1]])),
+    ],
+)
+def test_simpson_binary_distance_against_threshold(
+    vec_a,
+    vec_b,
```
j-adamczyk, line 36: Remove the comma, this can be in one line

j-adamczyk (changes requested):

skfp/distances/simpson.py
StaniszewskiA: It's probably safe to assume that this decorator is not needed here. I copy-pasted it from Rand, where we use prange. I'll check the speedup nevertheless, but I agree that it prbably should not be here.

j-adamczyk (approved):

PR #396, opened 2025-02-12
Refactor distances and add T4 similarity

j-adamczyk:
## Changes

Refactored distances and tests to remove code duplication and unify the code. Also added CT4 similarity and distance.


my-alaska (dismissed):

my-alaska (approved):

skfp/distances/ct4.py
```diff
@@ -0,0 +1,381 @@
...
+@njit(parallel=True)
+def _ct4_count_numpy(vec_a: np.ndarray, vec_b: np.ndarray) -> tuple[float, float]:
+    vec_a = vec_a.astype(np.float64).ravel()
+    vec_b = vec_b.astype(np.float64).ravel()
+
+    dot_ab = 0.0
+    dot_aa = 0.0
+    dot_bb = 0.0
+
+    for i in prange(vec_a.shape[0]):
```
mjste, line 362: Just wondering if its faster than simply 3 numpy dot products

skfp/distances/dice.py
```diff
@@ -361,12 +349,12 @@ def _dice_count_numpy(vec_a: np.ndarray, vec_b: np.ndarray) -> float:
         dot_aa += vec_a[i] * vec_a[i]
         dot_bb += vec_b[i] * vec_b[i]
 
-    return 2 * dot_ab / (dot_aa + dot_bb)
+    return float(2 * dot_ab / (dot_aa + dot_bb))
```
mjste, line 352: I see pattern with njit calculating three dot products an outputting slightly changed variation of those. Maybe we should create one function which returns three dot product for many distances?

skfp/distances/simpson.py
```diff
@@ -69,38 +69,32 @@ def simpson_binary_similarity(
...
-    if np.sum(vec_a) == 0 == np.sum(vec_b):
+    if np.sum(vec_a) == 0 or np.sum(vec_b) == 0:
         return 0.0
 
-    if isinstance(vec_a, csr_array) and isinstance(vec_b, csr_array):
-        vec_a_bool = vec_a.astype(bool)
-        vec_b_bool = vec_b.astype(bool)
-        return _simpson_binary_scipy(vec_a_bool, vec_b_bool)
+    if isinstance(vec_a, np.ndarray):
+        num_common = np.sum(vec_a & vec_b)
```
mjste, line 90: maybe use `np.logical_and` instead of `&`

mjste (changes requested):
Maybe we should extract those njit fragments to single function?

skfp/distances/braun_blanquet.py
```diff
@@ -65,38 +65,32 @@ def braun_blanquet_binary_similarity(
...
     if np.sum(vec_a) == 0 == np.sum(vec_b):
-        return 0.0
+        return 1.0
 
-    if isinstance(vec_a, csr_array) and isinstance(vec_b, csr_array):
-        vec_a_bool = vec_a.astype(bool)
-        vec_b_bool = vec_b.astype(bool)
-        return _braun_blanquet_binary_scipy(vec_a_bool, vec_b_bool)
+    if isinstance(vec_a, np.ndarray):
+        num_common = np.sum(vec_a & vec_b)
```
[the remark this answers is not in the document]
j-adamczyk: done

skfp/distances/ct4.py
```diff
@@ -0,0 +1,381 @@
...
+    if isinstance(vec_a, np.ndarray):
+        intersection = np.sum(vec_a & vec_b)
+        union = np.sum(vec_a | vec_b)
+    else:
+        intersection = vec_a.multiply(vec_b).sum()
+        union = vec_a.sum() + vec_b.sum() - intersection
+
+    if intersection in {0, 1} or union in {0, 1}:
+        # log of 0 is -infinity, and log of 1 is 0
+        return 0.0
```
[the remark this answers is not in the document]
j-adamczyk: Because that was the simplest one, and there is no definition for CT4 for this case. Here we can have 0/0, 0/inf, inf/0, and inf/inf. I decided to assume zero for all. Do you think that 1 makes more sense?

skfp/distances/ct4.py
j-adamczyk: It's comparable, I rewrote this

skfp/distances/simpson.py
j-adamczyk: done

mjste (approved):

PR #414, opened 2025-03-04
Add Bulk Rand similarity

StaniszewskiA:
## Changes

- [X] Docstrings added/updated in public functions and classes
- [X] Sphinx docs added/updated and render properly (`make docs` and see `docs/_build/index.html`)

### DIFF
--- a/skfp/distances/rand.py
+++ b/skfp/distances/rand.py
@@ -1,5 +1,6 @@
-from typing import Union
+from typing import Optional, Union
 
+import numba
 import numpy as np
 from scipy.sparse import csr_array
 from sklearn.utils._param_validation import validate_params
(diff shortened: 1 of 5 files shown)


skfp/distances/rand.py
```diff
@@ -166,3 +167,167 @@ def rand_binary_distance(
...
+def _bulk_rand_binary_similarity_single(X: np.ndarray) -> np.ndarray:
+    m = X.shape[0]
+    sims = np.empty((m, m))
+
+    # upper triangle - actual similarities
+    for i in numba.prange(m):
+        # in this case diagonal is not always 1
+        for j in numba.prange(i, m):
+            intersection = np.sum(np.logical_and(X[i], X[j]))
+            length = len(X[i])
```
j-adamczyk, line 251: Measure if a `for` loop for a diagonal would be faster. Diagonal can avoid using `np.logical_and`, since it can just sum up the vector

skfp/distances/rand.py
```diff
@@ -166,3 +167,167 @@ def rand_binary_distance(
...
+            sims[i, j] = intersection / length if length != 0 else 1.0
```
j-adamczyk, line 252: Don't check for zero length. Such data is nonsensical, and it's user's responsibility then

skfp/distances/rand.py
```diff
@@ -166,3 +167,167 @@ def rand_binary_distance(
...
+def _bulk_rand_binary_similarity_two(X: np.ndarray, Y: np.ndarray) -> np.ndarray:
+    m = X.shape[0]
+    n = Y.shape[0]
+    sims = np.empty((m, n))
+
+    for i in numba.prange(m):
+        for j in numba.prange(m):
+            intersection = np.sum(np.logical_and(X[i], Y[j]))
+            length = len(X[i])
+            sims[i, j] = intersection / length if length != 0 else 1.0
```
j-adamczyk, line 272: Don't check for length

j-adamczyk (changes requested):

PR #423, opened 2025-03-10
Bulk Rogot-Goldberg dist and sim

StaniszewskiA:
## Changes

- [X] Docstrings added/updated in public functions and classes
- [X] Sphinx docs added/updated and render properly (`make docs` and see `docs/_build/index.html`)

### DIFF
--- a/skfp/distances/harris_lahey.py
+++ b/skfp/distances/harris_lahey.py
@@ -280,16 +280,15 @@ def _bulk_harris_lahey_binary_similarity_single(
 ) -> np.ndarray:
     m = X.shape[0]
     sims = np.empty((m, m))
-    X_neg = 1 - X
 
     # upper triangle - actual similarities
     for i in numba.prange(m):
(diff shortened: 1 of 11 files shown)


skfp/distances/harris_lahey.py
```diff
@@ -280,16 +280,16 @@ def _bulk_harris_lahey_binary_similarity_single(
 ) -> np.ndarray:
     m = X.shape[0]
     sims = np.empty((m, m))
+    X_neg = 1 - X
```
j-adamczyk, line 283: You are doubling memory usage here. I would prefer to avoid this

j-adamczyk: Same in other places - this is already O(n^2) in memory due to `np.empty((m, m))`, and adding another O(n) here is a bad idea

j-adamczyk (changes requested):

j-adamczyk (approved):

PR #470, opened 2025-07-16
Klekota-Roth prefix optimization

Kacper-Kozubowski:
## Changes

- Completely rewrote the Klekota-Roth fingerprint computation for better performance.
- Replaced the naive O(n × m) loop over molecules and patterns with a tree-based traversal.
- Generated an efficient tree structure used by the algorithm and stored it in `tree.json`.
- Added atom-count-based filtering for faster pattern rejection.
- Removed the hardcoded list of 4860 patterns from the module; patterns are now loaded from the tree structure.
- Added a test verifying result correctness against the naive approach.

## Other notes
- `tree.json` is generated offline with a separate script (not included in this PR).
- Achieved a 6–9× performance boost for single-core execution and roughly a 5–7× boost for multicore, depending on dataset.


skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+
+    def _load_tree(self) -> None:
+        """
+        Load the pattern tree from a JSON file into internal representation.
+        """
+        file = _TREE_PATH
+        if not file.exists():
+            raise FileNotFoundError(f"Tree file not found: {file}")
+
+        with file.open("r", encoding="utf-8") as f:
```
j-adamczyk, line 216: Prefer `file` instead of `f`, one-letter variables are not a good practice

skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+            data = json.load(f)
+
+        for key in ("n_nodes", "atoms", "tree"):
+            if key not in data:
+                raise KeyError(f"Missing key {key} in tree file {file}")
```
j-adamczyk, line 221: We provide the file, I don't think this check is necessary

skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+
+        self._feature_names = [""] * data["n_terminal_nodes"]
+        self._pattern_atoms = {key: Chem.MolFromSmarts(key) for key in data["atoms"]}
+        self._root = self._dict_to_node(data["tree"])
+
+        if any(f == "" for f in self._feature_names):
+            raise ValueError("SMARTS or feature_bit missing in terminal nodes")
```
j-adamczyk, line 228: As above

skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+        node.feature_bit = d.get("feature_bit")
+        node.atom_requirements = defaultdict(int, d.get("atom_requirements", {}))
+        node.children = [
+            self._dict_to_node(node_dict) for node_dict in d.get("children", [])
+        ]
+
+        if (
+            node.is_terminal
+            and node.smarts is not None
+            and node.feature_bit is not None
```
j-adamczyk, line 203: How can the node be terminal and not have SMARTS or feature bit?

skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+        ):
+            self._feature_names[int(node.feature_bit)] = node.smarts
+        return node
```
j-adamczyk, line 206: I prefer to have 1 empty line between `if` or `for` and further code for clarity

skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+        """
+        Recursively convert a dict representation of a pattern tree
+        into a _PatternNode tree.
+        """
+        node = _PatternNode()
+        node.smarts = d.get("smarts")
+        node.pattern_mol = Chem.MolFromSmarts(node.smarts) if node.smarts else None
+        node.is_terminal = d.get("is_terminal", False)
+        node.feature_bit = d.get("feature_bit")
+        node.atom_requirements = defaultdict(int, d.get("atom_requirements", {}))
```
j-adamczyk, line 195: If there are atom requirements, they should be specified explicitly, right? So why `defaultdict` with default value 0?

skfp/fingerprints/klekota_roth.py
```diff
@@ -4990,3 +181,82 @@ def transform(
...
+        if self.count:
+            set_value = lambda mol, pattern: len(mol.GetSubstructMatches(pattern))
+        else:
+            set_value = lambda _mol, _pattern: 1
+
+        for i, mol in enumerate(X):
+            stack: deque[_PatternNode] = deque(root_children)
+            atom_contents = defaultdict(int)
+            for key, atom in self._pattern_atoms.items():
+                atom_contents[key] = len(mol.GetSubstructMatches(atom))
```
j-adamczyk, line 246: For checking atom types, iteration over atoms should be faster than individual SMARTS patterns

tests/fingerprints/klekota_roth.py
```diff
@@ -50,3 +54,30 @@ def test_klekota_roth_feature_names():
 
     assert len(feature_names) == kr_fp.n_features_out
     assert len(feature_names) == len(set(feature_names))
+
+
+@pytest.mark.parametrize("count", [True, False])
```
j-adamczyk, line 59: `False, True` to keep convention of other tests

j-adamczyk (approved):

Issue #459, opened 2025-06-28
Add condition names for filters


j-adamczyk:
Currently, there is no way to inspect the detailed results of filters, e.g. Brenk or ZINC, or names of physicochemical properties for other filters. The interpretable descriptions of substructures are available as attributes of `FilterCatalog`, and should be read and saved as a filter attribute, similarly for physicochemical properties. This also requires exposing results of individual matches, so we could either make another parameter (e.g. `return_match_indicators`), or change `return_indicators` to something like `return_type` with values `molecules`, `mol_indicators` and `match_indicators`.

PR #475, opened 2025-07-27
Substructural filters conditions

j-adamczyk:
## Changes

Adds condition names for substructural filters, closes https://github.com/scikit-fingerprints/scikit-fingerprints/issues/459. Required a bit of refactoring and introducing a new base class, but also reduced a lot of repeated code in tests.


skfp/bases/base_filter.py
```diff
@@ -34,6 +34,10 @@ class BaseFilter(ABC, BaseEstimator, TransformerMixin):
 
     Parameters
     ----------
+    condition_names : list[str]
+        Names of filters conditions, e.g. physicochemical properties and their limits,
```
my-alaska, line 38: "filter conditions" or "filter's conditions" instead of "filters conditions"

skfp/filters/pains.py
```diff
@@ -85,36 +101,35 @@ def __init__(
...
+        elif self.variant == "B":
             filter_rules = FilterCatalogParams.FilterCatalogs.PAINS_B
-        elif variant == "C":
+        elif self.variant == "C":
             filter_rules = FilterCatalogParams.FilterCatalogs.PAINS_C
         else:
-            raise ValueError(f'PAINS variant must be "A", "B" or "C", got {variant}')
+            raise ValueError(
+                f'PAINS variant must be "A", "B" or "C", got {self.variant}'
+            )
```
my-alaska, line 130: Consider raising this error during model initialization to catch it earlier

skfp/bases/base_filter.py
j-adamczyk: Right

skfp/filters/pains.py
j-adamczyk: It is, `"variant": [StrOptions({"A", "B", "C"})],` is defined in parametr constraints. Here it's just another check just in case. But I will change the error type for consistency.

Issue #424, opened 2025-03-10
Implement applicability domain algorithms


j-adamczyk:
Applicability domain checks whether new molecules are inside the densely sampled space near the training set. The idea is that the farther out is a new molecule, the harder it is to predict it. It's important when internal evaluation splits like random or MaxMin are used. There are a few algorithms to check it, see e.g.:
- [Applicability Domain: A Step Toward Confident Predictions and Decidability for QSAR Modeling](https://annas-archive.org/scidb/10.1007/978-1-4939-7899-1_6/)
- ["Comparative Evaluation of Applicability Domain Definition Methods for Regression Models"](https://arxiv.org/pdf/2411.00920)
- ["Efficiency of different measures for defining the applicability domain of classification models"](https://jcheminf.biomedcentral.com/articles/10.1186/s13321-017-0230-2)
- ["Applicability Domain: Towards a more formal definition"](https://bigchem.eu/sites/default/files/Thierry_Hanser.pdf)
- ["Applicability domain of QSAR models: status quo and perspectives"](https://cadaster.eu/sites/cadaster.eu/files/workshop/Sushko.pdf)
- [MLChemAD library](https://github.com/OlivierBeq/MLChemAD)

Implementation notes:
- we should treat this similarly to filters, i.e. either filter out molecules not in AD or flag them as 0/1 (0 - outside AD)
- inputs are most often molecular fingerprints
- we should not scale or preprocess inside those classes in most cases, user should do this himself, but documentation should mention this

Algorithms to implement:
- [x] bounding box
- [x] PCA bounding box
- [x] convex hull
- [x] distance to centroids
- [x] Hotelling T^2 test
- [x] kNN
- [x] leverage
- [x] PROBSTD (see 3rd reference above)
- [ ] TOPKAT
- [ ] response variable range
- [ ] standard deviation

PR #479, opened 2025-08-06
Add ResponseVariableRange and StdEnsemble AD checkers with tests

Kacper-Kozubowski:
## Changes

Added `ResponseVariableRange` and `StdEnsemble` applicability domain (AD) checkers, and provided tests.
Also, added missing references to the documentation of the `BoundingBoxADChecker`. Part of  https://github.com/scikit-fingerprints/scikit-fingerprints/issues/424

### DIFF
--- a/skfp/applicability_domain/response_variable_range.py
+++ b/skfp/applicability_domain/response_variable_range.py
@@ -1,0 +1,98 @@
+import numpy as np
+from sklearn.utils.validation import check_is_fitted, validate_data
+
+from skfp.bases.base_ad_checker import BaseADChecker
+
+
+class ResponseVariableRangeADChecker(BaseADChecker):
+    """
+    Response variable range method.
+
+    Defines applicability domain based on the range of response values observed
+    in the training data. New predictions are considered inside the applicability
+    domain if they lie within the min-max range of training targets.
+
+    Typically, this method is used after model prediction, and checks whether
+    predicted values lie within the known domain of the response variable.
+
+    Note that this method does not consider molecular structure or descriptors.
+    It operates purely in the output (target) space.
+
+    This method scales extremely well with both number of samples and features.
+
+    Parameters
+    ----------
+    n_jobs : int, default=None
+        The number of jobs to run in parallel. :meth:`transform_x_y` and
+        :meth:`transform` are parallelized over the input molecules. ``None`` means 1
+        unless in a :obj:`joblib.parallel_backend` context. ``-1`` means using all
+        processors. See scikit-learn documentation on ``n_jobs`` for more details.
+
+    verbose : int or dict, default=0
+        Controls the verbosity when filtering molecules.
+        If a dictionary is passed, it is treated as kwargs for ``tqdm()``,
+        and can be used to control the progress bar.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from skfp.applicability_domain import ResponseVariableRangeADChecker
+    >>> y_train = np.array([0.5, 1.2, 1.5, 0.9])
+    >>> y_pred = np.array([1.0, 1.6, 0.4])
+    >>> response_range_ad_checker = ResponseVariableRangeADChecker()
+    >>> response_range_ad_checker
+    ResponseVariableRangeADChecker()
+
+    >>> response_range_ad_checker.fit(y_train)
+    ResponseVariableRangeADChecker()
+
+    >>> response_range_ad_checker.predict(y_pred)
+    array([ True, False, False])
+    """
+
+    def __init__(
+        self,
+        n_jobs: int | None = None,
+        verbose: int | dict = 0,
+    ):
+        super().__init__(
+            n_jobs=n_jobs,
+            verbose=verbose,
+        )
+
+    def fit(  # noqa: D102
+        self,
+        X: np.ndarray,
+        y: np.ndarray | None = None,
+    ):
+        y = validate_data(self, X=X, ensure_2d=False)
+        self.min_y_ = np.min(y)
+        self.max_y_ = np.max(y)
+        self.mean_y_ = np.mean(y)
+        return self
+
+    def predict(self, y: np.ndarray) -> np.ndarray:  # noqa: D102
+        check_is_fitted(self)
+        y = validate_data(self, X=y, ensure_2d=False, reset=False)
+        return (self.min_y_ <= y) & (self.max_y_ >= y)
+
+    def score_samples(self, y: np.ndarray) -> np.ndarray:
+        """
+        Calculate the applicability domain score of samples. It is defined as
+        the absolute distance of each predicted value from the training response mean.
+        Lower scores indicate that a prediction is closer to the center of the training
+        response distribution.
+
+        Parameters
+        ----------
+        y : array-like of shape (n_samples,)
+            Predicted response values.
+
+        Returns
+        -------
+        scores : ndarray of shape (n_samples,)
+            Applicability domain scores of samples.
+
+        """
+        y = validate_data(self, X=y, ensure_2d=False, reset=False)
+        return np.abs(y - self.mean_y_)
--- a/skfp/applicability_domain/response_variable_range.py
+++ b/skfp/applicability_domain/response_variable_range.py
@@ -9,7 +9,7 @@ class ResponseVariableRangeADChecker(BaseADChecker):
     Response variable range method.
 
     Defines applicability domain based on the range of response values observed
-    in the training data. New predictions are considered inside the applicability
+    in the training data [1]_. New predictions are considered inside the applicability
     domain if they lie within the min-max range of training targets.
 
     Typically, this method is used after model prediction, and checks whether
@@ -18,7 +18,8 @@ class ResponseVariableRangeADChecker(BaseADChecker):
     Note that this method does not consider molecular structure or descriptors.
     It operates purely in the output (target) space.
 
-    This method scales extremely well with both number of samples and features.
+    This method scales extremely well with the number of samples,
+    as it only operates in the 1D target space.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


skfp/applicability_domain/response_variable_range.py
```diff
@@ -0,0 +1,108 @@
...
+
+    >>> response_range_ad_checker.fit(y_train)
+    ResponseVariableRangeADChecker()
+
+    >>> response_range_ad_checker.predict(y_pred)
+    array([ True, False, False])
+    """
+
+    def __init__(
+        self,
```
j-adamczyk, line 64: Introduce optional `threshold` parameter. If `None` (default), we use min and max as thresholds, as is now. If it's a float, we use `[mean - threshold, mean + threshold]` as allowed range

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
+from numbers import Real
+
+import numpy as np
+from sklearn.utils._param_validation import Interval
+from sklearn.utils.validation import validate_data
+
+from skfp.bases.base_ad_checker import BaseADChecker
+
+
+class StdEnsembleADChecker(BaseADChecker):
```
j-adamczyk, line 10: `StandardDeviationADChecker` - since standard deviation has to come from an ensemble, it's kind of obvious

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Standard deviation ensemble method.
```
j-adamczyk, line 12: Without `ensemble`

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Defines applicability domain based on the spread (standard deviation) of predictions
+    from individual estimators in an ensemble model [1]_. The method assumes that
+    if different estimators give similar predictions for a given molecule,
+    the prediction is more likely to be reliable (i.e., in-domain).
+
+    This approach requires an ensemble model exposing the ``estimators_``
+    attribute (e.g., RandomForestRegressor), where each
+    sub-model must implement a ``predict(X)`` method. At prediction time,
```
j-adamczyk, line 21: ``.predict(X)`` (just to signify that it is a method)

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
...
+    each sample is passed to all estimators, and the standard deviation of
+    their predictions is calculated. The sample is considered in-domain if
+    the standard deviation is lower than or equal to the specified threshold.
+
+    Typically, physicochemical properties (continous features) are used as inputs.
+    Consider scaling, normalizing, or transforming them before computing AD to lessen
+    effects of outliers, e.g. with ``PowerTransformer`` or ``RobustScaler``.
```
j-adamczyk, line 28: Here it's not necessary, e.g. model on ECFP fingerprints would work equally well

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    This method is particularly useful in assessing uncertainty in ensemble
+    learning-based QSAR models.
```
j-adamczyk, line 31: Merge this with the first paragraph, since here you basically say `This method assesses model uncertainty, linked to the model reliability, which is inherently higher in densely sampled regions inside training data.`

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Parameters
+    ----------
+    model : object
+        Fitted ensemble model with accessible ``estimators_`` attribute and
+        ``predict(X)`` method on each sub-estimator.
```
j-adamczyk, line 37: `.predict(x)`

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,137 @@
...
+        self.threshold = threshold
+
+    def fit(  # noqa: D102
+        self,
+        X: np.ndarray,  # noqa: ARG002
+        y: np.ndarray | None = None,  # noqa: ARG002
+    ):
+        return self
+
+    def _predict_all_estimators(self, X: np.ndarray) -> np.ndarray:
```
j-adamczyk, line 108: Implement this below `.predict()` - it's better to have public methods before private methods

j-adamczyk (changes requested):

tests/applicability_domain/response_variable_range.py
```diff
@@ -0,0 +1,68 @@
...
+    assert isinstance(preds, np.ndarray)
+    assert np.isdtype(preds.dtype, np.bool)
+    assert preds.shape == (len(y_test),)
+    assert np.all(preds == 0)
+
+
+def test_response_range_with_threshold():
+    y_train = np.array([0.8, 0.9, 1.0, 1.1, 1.2])
+    y_test_in = np.array([0.95, 1.0, 1.05])
+    y_test_out = np.array([0.85, 1.15])
```
my-alaska, line 46: We should test that relevant error with proper information is thrown if the user passes multilabel data. Alternatively add multilabel support in response variable range AD. Would require defining logic for it and additional tests

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,130 @@
...
+
+    _parameter_constraints: dict = {
+        **BaseADChecker._parameter_constraints,
+        "model": [object],
+        "threshold": [Interval(Real, 0, None, closed="left")],
+    }
+
+    def __init__(
+        self,
+        model,
```
my-alaska, line 82: I think we need to consider 3 scenarios

- User passes a trained model
- User passes data that is then used to fit a default model inside the constructor
- User passes data and a fresh model that is fitted to the data inside the constructor

Right now only the first scenario is handled. Consider adding the other two as user might want a default to work with. It will also allow for neater code

my-alaska (changes requested):

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Standard deviation method.
+
+    Defines applicability domain based on the spread (standard deviation) of predictions
+    from individual estimators in an ensemble model [1]_. The method assumes that
+    greater consensus among estimators indicates higher prediction reliability,
+    which typically occurs in densely sampled regions of the training data.
+
+    This approach requires an ensemble model exposing the ``estimators_``
+    attribute (e.g., RandomForestRegressor), where each
```
j-adamczyk, line 20: Add backticks for model name. Also, I had an idea now that classifiers should also work, as long as they define a `.predict_proba()` method, we should support that. After all, probability estimates also can have standard deviation

j-adamczyk (changes requested):

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,154 @@
...
+        X = validate_data(self, X=X, reset=False)
+        check_is_fitted(self.model, "estimators_")
+
+        preds = self._predict_all_estimators(X)
+        stds = np.std(preds, axis=1)
+
+        return (stds <= self.threshold).astype(bool)
+
+    def _predict_all_estimators(self, X: np.ndarray) -> np.ndarray:
+        if hasattr(self.model.estimators_[0], "predict_proba"):  # type: ignore[union-attr]
```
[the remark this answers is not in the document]
j-adamczyk: Ok, makes sense, you can leave it here then

skfp/applicability_domain/standard_deviation.py
```diff
@@ -0,0 +1,154 @@
...
+            preds = np.array([est.predict_proba(X) for est in self.model.estimators_])  # type: ignore[union-attr]
+            if preds.shape[2] == 2:
+                preds = preds[:, :, 1]  # shape: (n_estimators, n_samples)
```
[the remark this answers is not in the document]
Kacper-Kozubowski: Here we need the size of the last dimension, not just the number of dimensions, so `.ndim` wouldn’t work. At this point, `preds` will always have 3 dimensions `(n_estimators, n_samples, n_classes)`, and the `n_classes` value is what matters for us.

skfp/applicability_domain/response_variable_range.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+        self.threshold = threshold
+
+    def fit(  # type: ignore[override]  # noqa: D102
+        self,
+        y: np.ndarray,
+        X: np.ndarray | None = None,  # noqa: ARG002
+    ):
+        y = check_array(y, ensure_2d=False, allow_nd=False)
+        if y.ndim != 1:
```
j-adamczyk, line 99: This should be covered by `allow_nd=False`, right?

skfp/applicability_domain/response_variable_range.py
```diff
@@ -0,0 +1,151 @@
...
+    def predict(self, y: np.ndarray) -> np.ndarray:  # noqa: D102
+        check_is_fitted(self)
+        y = check_array(y, ensure_2d=False, allow_nd=False)
+        if y.ndim != 1:
+            raise ValueError(
+                f"{self.__class__.__name__} only supports 1D target values, "
+                f"but got y with shape {y.shape}."
+            )
+
+        return (self.lower_bound_ <= y) & (self.upper_bound_ >= y)
```
j-adamczyk, line 125: If those are floats, you can just do `self.lower_bount <= y <= self.upper_bound_`

skfp/applicability_domain/standard_deviation.py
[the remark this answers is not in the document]
j-adamczyk: You can combine check for `predict_proba` attribute, `is_classifier` function (https://scikit-learn.org/stable/modules/generated/sklearn.base.is_classifier.html), and check `.classes_` length (attribute created by all classifiers). Also, do this check once in `validate_params()`, rather than here. After all, if user provides wrong parameter, it should raise `InvalidParameterError`

j-adamczyk (changes requested):

skfp/applicability_domain/response_variable_range.py
Kacper-Kozubowski: Chained comparisons don't seem to work with numpy arrays

PR #561, opened 2026-05-01
fix: return arrays for parallel filter indicators

### LINKED ISSUE #557: Filter indicators wrong return type

Genmin:
## Summary
- normalize non-condition filter indicators to a NumPy boolean array after the parallel path
- add regression coverage for parallel indicator output type

Fixes #557

## Validation
- uv run pytest tests/bases/base_filter.py -q
- uv run ruff check skfp/bases/base_filter.py tests/bases/base_filter.py
- uv run ruff format --check skfp/bases/base_filter.py tests/bases/base_filter.py
- git diff --check


j-adamczyk:
Closed to due suspected bot activity.

PR #565, opened 2026-05-01
Refactor LINGO and tests

j-adamczyk:
## Changes

Covers https://github.com/MLCIL/scikit-fingerprints/issues/552, refactored & sped up LINGO fingerprint. Also changed implementation to be actually more accurate compared to the paper, thanks to Andrew Dalke.


tests/fingerprints/lingo.py
```diff
@@ -1,113 +1,90 @@
...
-    if "tests" in os.listdir():
-        return loader(os.path.join("tests", "fingerprints", "data", filename))
-    if "fingerprints" in os.listdir():
-        return loader(os.path.join("fingerprints", "data", filename))
-    if "data" in os.listdir():
-        return loader(os.path.join("data", filename))
-
-    raise FileNotFoundError(f"File {filename} not found")
+def test_lingo_bit_fingerprint(smiles_list, mols_list):
+    lingo_fp_seq = LingoFingerprint(n_jobs=-1)
```
mjste: Jobs=**+**1

tests/fingerprints/lingo.py
```diff
@@ -1,113 +1,90 @@
...
+    X_parallel = lingo_fp_parallel.transform(smiles_list)
+
+    assert_equal(X_seq, X_parallel)
+    assert_equal(X_seq.shape, (len(smiles_list), lingo_fp_seq.fp_size))
+    assert X_seq.dtype == np.uint8
+    assert np.all(np.isin(X_seq, [0, 1]))
+
+
+def test_lingo_count_fingerprint(smiles_list, mols_list):
+    lingo_fp_seq = LingoFingerprint(count=True, n_jobs=-1)
```
mjste: Ditto

mjste (changes requested):

skfp/fingerprints/lingo.py
```diff
@@ -128,54 +129,47 @@ def transform(
...
-        return result
+        # replace numbers for rings, but not charges in []
+        # we use negative lookback/lookahead in regex
+        X = [re.sub(r"(?<!\[)%?\d+(?!\])", "0", smi) for smi in X]
+        X = [
+            smi.replace("Cl", "L").replace("Br", "R")
+            if "Cl" in smi or "Br" in smi
+            else smi
+            for smi in X
+        ]
```
mjste: Why not just replace? Is there any significant speedup for checking for substring first?

tests/fingerprints/lingo.py
```diff
@@ -1,113 +1,90 @@
...
+            "c0cc0cc0c0c0c(cc0ccc0cc0cc0c0c0c0c0c0c0c0c0c0c(c0C0N0CCOCCOCCOCCOCCOCC0)c0c0)C0",
+            True,
+        ),
+        # common 2-letter elements simplification: Cl -> L, Br -> R
+        ("[Na+].[Cl-]", "[Na+].[L-]", True),
+        ("ClN(Cl)Cl", "LN(L)L", True),
+        ("BrC#N", "RC#N", True),
+        ("BrCl", "RL", True),
+        # charges in brackets [] shouldn't be touched and give different fingerprints
+        ("[Cl+3]", "Cl+2]", False),
```
mjste: Typo? Missing bracket

tests/fingerprints/lingo.py
```diff
@@ -1,113 +1,90 @@
...
+        ("[12CH0]OOOO[Al-2]", "[12CH2]OOOO[Al-0]", False),
```
mjste, line 80: I see two changes here. Is it on purpose?

skfp/fingerprints/lingo.py
```diff
@@ -128,54 +129,47 @@ def transform(
...
+        X = self._smiles_to_dicts(X)
+        X = self._dicts_to_array(X)
+        return csr_array(X) if self.sparse else X
 
-        Returns
-        -------
-        result: list[dict[str, int]]
-            List of dictionaries containing substring counts.
-        """
+    def _smiles_to_dicts(self, X: Sequence[str | Mol]) -> list[dict[str, int]]:
```
mjste, line 137: This was public earlier, is it on purpose?

skfp/fingerprints/lingo.py
j-adamczyk: It turns out that the speedup isn't statistically significant when I bechmarked it more thoroughly, I removed this

tests/fingerprints/lingo.py
j-adamczyk: good catch

tests/fingerprints/lingo.py
j-adamczyk: Yes, both shouldn't be normalized, so there is no equality

skfp/fingerprints/lingo.py
j-adamczyk: Yes, I thought this shouldn't be public really. I see no reason why anyone would like to use raw dictionaries here. This is a kind of method that we should be able to change as we want internally. But this is arguable.

j-adamczyk:
Added all changes from review

mjste (approved):

PR #605, opened 2026-05-30
Mordred autocorrelation

j-adamczyk:
## Changes

Mordred autocorrelation descriptors. ~3.5x faster than original.


tests/fingerprints/_new_mordred/test_sasa.py
OlaszPL, line 1: Previously, we decided to use the SASA implementation from rdkit instead of the manual one. It only passed the reference test from mordred-community when using the mass/vdw_radii values from rdkit (I don't remember which exactly). Using the values from the `.txt` files caused the results to diverge (and now the pipeline fails on this test). I know that the values differ but we should investigate whether mass/vdw_radii is only used in SASA or to use rdkit ones in `sasa.py`.

skfp/fingerprints/_new_mordred/utils/periodic_table.py
OlaszPL, line 1: As said before - does autocorrelation gives different values when rdkit mass/vdw_radii is used? If so, maybe try to use rdkit ones in SASA? To pass the reference test?

j-adamczyk: Yes, it does. The reference test uses mass & van der Waals radii from PaDEL-Descriptor, which are different from RDKit

tests/fingerprints/_new_mordred/test_sasa.py
j-adamczyk: I added a separate function, so we can have both

skfp/fingerprints/_new_mordred/calculator.py
```diff
@@ -48,10 +47,15 @@ def compute(mol: Mol, use_3D: bool) -> np.ndarray:
     # dependencies
     n_frags = len(GetMolFrags(mol))
 
+    # classic, RDKit-standardized molecule
     mol_regular = preprocess_mol(mol)
     distance_matrix_regular = DistanceMatrix(mol_regular)
     adjacency_matrix_regular = AdjacencyMatrix(mol_regular)
 
+    # hydrogen-explicit moleculr
```
OlaszPL:
```suggestion
    # hydrogen-explicit molecule
```

OlaszPL (approved):
LGTM 🚀

FloudMe77 (approved):

Issue #328, opened 2024-11-15
Implement molecular clustering


j-adamczyk:
Currently, splitters used custom clustering implementations:
- MaxMin by diversity and nearest neighbors
- Butina by leader following clustering
- scaffold by scaffold grouping

We should implement this as proper scikit-learn clustering classes. This won't be a lot of new code, since it's already implemented, rather we will move the existing code.

Clusterings to implement:
- [ ] MaxMin
- [ ] Butina (leader picker / leader following)
- [ ] scaffold

rayair250-droid:
I'd like to take the Butina part of this. I'll add a `ButinaClustering` estimator mirroring the existing `MaxMinClustering` (`BaseEstimator`/`ClusterMixin`, `fit`/`predict`, `_parameter_constraints`, `labels_`), reusing the clustering logic currently in `butina_split.py`, with tests matching the MaxMin clustering tests. Will keep the splitter working (it can delegate to the new class in a follow-up if you prefer).

PR #625, opened 2026-07-03
Add ButinaClustering estimator

rayair250-droid:
Part of #328 (Butina). Adds a scikit-learn-style `ButinaClustering` estimator, mirroring the existing `MaxMinClustering`.

### What it does
- `ButinaClustering(BaseEstimator, ClusterMixin)` with `_parameter_constraints`, `fit`, `predict`, `fit_predict`, `get_clusters_and_points`, and the same `centroid_indices_` / `centroid_bitvectors_` / `centroids_` / `labels_` attributes as `MaxMinClustering`.
- Clustering uses RDKit `Butina.ClusterData` on condensed lower-triangle Tanimoto distances (the RDKit-documented `isDistData=True` format), with `reordering=True` for deterministic output.
- Accepts the same inputs as `MaxMinClustering`: dense NumPy arrays, CSR sparse arrays, or sequences of `ExplicitBitVect`.
- `predict` assigns new samples to the nearest centroid by Tanimoto similarity.

### Notes / design choices to confirm
- I kept this to the new estimator only; the existing `butina_split` splitter is untouched (it can delegate to this class in a follow-up if you prefer, per the issue).
- `labels_` reflect Butina cluster membership (density/sphere-exclusion), which is not necessarily the nearest-centroid assignment — so unlike the MaxMin tests, I did not assert nearest-centroid for `fit` labels. `predict` (new samples) does use nearest-centroid.

### Verification
`ruff check` + `ruff format --check` clean; `pytest tests/clustering/` → **38 passed** (19 new for Butina, existing MaxMin unchanged). Verified against an editable install of the package.

Happy to adjust the API to match your intent for the scaffold/Butina classes.

### DIFF
--- a/skfp/clustering/butina.py
+++ b/skfp/clustering/butina.py
@@ -9,22 +9,20 @@ from sklearn.base import BaseEstimator, ClusterMixin
 from sklearn.utils._param_validation import Interval, RealNotInt
 from sklearn.utils.validation import check_is_fitted, validate_data
 
+from skfp.clustering import utils
+
 
 class ButinaClustering(BaseEstimator, ClusterMixin):
     """
     Taylor-Butina clustering.
 
     A density-based clustering algorithm for binary fingerprints using Tanimoto
-    similarity, also known as sphere exclusion or leader-following clustering [1]_ [2]_.
-    Cluster centroids are chosen so that no two centroids are closer than a given
-    Tanimoto distance ``distance_threshold``; every remaining sample joins the cluster
-    of the first centroid within that radius. In contrast to centroid-based methods
-    like MaxMin clustering, Butina clusters follow the density of the data and can vary
-    widely in size.
-
-    Clustering uses RDKit ``Butina.ClusterData`` with ``reordering=True`` for
-    deterministic output. After fitting, new samples are assigned with :meth:`predict`
-    to the centroid with the highest Tanimoto similarity.
+    similarity, also known as sphere exclusion clustering [1]_ [2]_. Cluster centroids
+    are chosen so that no two centroids are closer than a given Tanimoto distance
+    ``distance_threshold``; every remaining sample joins the cluster of the first
+    centroid within that radius. In contrast to centroid-based methods like MaxMin
+    clustering, Butina clusters follow the density of the data and can vary widely
+    in size.
 
     Parameters
     ----------
@@ -102,7 +100,7 @@ class ButinaClustering(BaseEstimator, ClusterMixin):
         super()._validate_params()
         X = validate_data(self, X, accept_sparse=["csr"], ensure_2d=False)
 
-        fps = self._array_to_bitvectors(X)
+        fps = utils.array_to_bitvectors(X)
         n_samples = len(fps)
 
         # Condensed lower-triangle Tanimoto distances, the format expected by
@@ -155,8 +153,8 @@ class ButinaClustering(BaseEstimator, ClusterMixin):
         """
         check_is_fitted(self)
         X = validate_data(self, X, accept_sparse=["csr"], ensure_2d=False)
-        bitvecs = self._array_to_bitvectors(X)
-        return self._assign_labels(bitvecs)
+        bitvecs = utils.array_to_bitvectors(X)
+        return utils.assign_labels(bitvecs, self.centroid_bitvectors_)
 
     def fit_predict(
         self, X: np.ndarray | sparse.csr_array | Sequence[ExplicitBitVect], y=None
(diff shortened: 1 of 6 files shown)


skfp/clustering/butina.py
```diff
@@ -0,0 +1,221 @@
...
+from sklearn.utils._param_validation import Interval, RealNotInt
+from sklearn.utils.validation import check_is_fitted, validate_data
+
+
+class ButinaClustering(BaseEstimator, ClusterMixin):
+    """
+    Taylor-Butina clustering.
+
+    A density-based clustering algorithm for binary fingerprints using Tanimoto
+    similarity, also known as sphere exclusion or leader-following clustering [1]_ [2]_.
```
j-adamczyk: I would avoid this naming, the leader-following has certain differences, e.g. see slide 24 in Roger Sayle's talk

skfp/clustering/butina.py
```diff
@@ -0,0 +1,221 @@
...
+    Cluster centroids are chosen so that no two centroids are closer than a given
+    Tanimoto distance ``distance_threshold``; every remaining sample joins the cluster
+    of the first centroid within that radius. In contrast to centroid-based methods
+    like MaxMin clustering, Butina clusters follow the density of the data and can vary
+    widely in size.
+
+    Clustering uses RDKit ``Butina.ClusterData`` with ``reordering=True`` for
+    deterministic output. After fitting, new samples are assigned with :meth:`predict`
+    to the centroid with the highest Tanimoto similarity.
```
j-adamczyk: Too much technical detail in docstring, remove

skfp/clustering/butina.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _assign_labels(self, vectors: list[ExplicitBitVect]) -> np.ndarray:
+        """
+        Assign each sample to the nearest centroid by Tanimoto similarity.
+        """
+        labels = np.empty(len(vectors), dtype=int)
+        for i, fp in enumerate(vectors):
+            sims = BulkTanimotoSimilarity(fp, self.centroid_bitvectors_)
+            labels[i] = int(np.argmax(sims))
+        return labels
```
j-adamczyk: 100% code duplication with MaxMin clustering. Create utils.py file in clustering and move those methods there as functions. Then also modify MaxMin clustering appropriately.

skfp/clustering/butina.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def get_clusters_and_points(self) -> dict[int, np.ndarray]:
+        """
+        Return a mapping from cluster id to the indices of its member samples.
+        """
+        check_is_fitted(self)
+        return {
+            k: np.where(self.labels_ == k)[0]
+            for k in range(len(self.centroid_indices_))
+        }
```
j-adamczyk: If you rewrite this to use `self.labels_` directly, this can be moved to the parent class and be common for all clustering classes. This mapping does not require any notion of centroids, after all.

tests/clustering/butina.py
```diff
@@ -0,0 +1,118 @@
...
+    dense = np.asarray(binary_X.todense()) if hasattr(binary_X, "todense") else binary_X
+    labels_dense = ButinaClustering(distance_threshold=0.5).fit(dense).labels_
+    labels_sparse = (
+        ButinaClustering(distance_threshold=0.5).fit(csr_matrix(dense)).labels_
+    )
+    assert np.array_equal(labels_dense, labels_sparse)
+
+
+def test_predict_assigns_to_nearest_centroid(binary_X):
+    from rdkit.DataStructs import BulkTanimotoSimilarity
```
j-adamczyk: Import globally

j-adamczyk (changes requested):

rayair250-droid:
Thanks for the detailed review! Addressed everything in the latest push:

1. Dropped the `leader-following` naming.
2. Trimmed the extra technical paragraph from the docstring.
3. Moved `array_to_bitvectors` and `assign_labels` into `skfp/clustering/utils.py` and refactored **both** `ButinaClustering` and `MaxMinClustering` to use them (removed the duplicated private methods).
4. `get_clusters_and_points` now derives from `labels_` directly via a shared `utils.clusters_and_points` helper. I kept it as a utils function called by both classes rather than introducing a new base class, to stay minimal — happy to move it to a shared base class if you'd prefer that.
5. Global imports in the tests.

All 38 clustering tests pass; `ruff check`/`ruff format` clean.

skfp/clustering/butina.py
```diff
@@ -0,0 +1,175 @@
...
+import numpy as np
+from rdkit.DataStructs import BulkTanimotoSimilarity
+from rdkit.DataStructs.cDataStructs import ExplicitBitVect
+from rdkit.ML.Cluster import Butina
+from scipy import sparse
+from sklearn.base import BaseEstimator, ClusterMixin
+from sklearn.utils._param_validation import Interval, RealNotInt
+from sklearn.utils.validation import check_is_fitted, validate_data
+
+from skfp.clustering import utils
```
j-adamczyk, line 12: Import directly from utils, this is the convention that we use

skfp/clustering/utils.py
```diff
@@ -0,0 +1,47 @@
+import numpy as np
+from rdkit.DataStructs import BulkTanimotoSimilarity
+from rdkit.DataStructs.cDataStructs import ExplicitBitVect
+from scipy import sparse
+
+
+def array_to_bitvectors(X) -> list[ExplicitBitVect]:
+    """Convert input data to a list of RDKit ExplicitBitVect objects."""
```
j-adamczyk, line 8: All docstrings use 3 separate lines, even oneliners, by convention

j-adamczyk (changes requested):

PR #626, opened 2026-07-03
Support pathlib.Path in MolFromSDFTransformer

### LINKED ISSUE #608: Support `Path` for file reading
Currently, file readers like `MolFromSDFTransformer` only support strings. We should also support `Path` from `pathlib`, but make sure that it works with RDKit underneath.

rayair250-droid:
Closes #608.

`MolFromSDFTransformer.transform` only accepted `str`. A `pathlib.Path` broke at `X.endswith(".sdf")` (Path has no `.endswith`).

Fix: normalize the input with `os.fspath(X)` at the start of `transform`, so a `Path` is accepted while `str` paths and raw SDF text pass through unchanged. Type hint and docstring widened to `str | Path`.

Scope: this addresses the reader named in the issue. `MolToSDFTransformer` takes its output path as a constructor param constrained to `[str]`; happy to extend that to `PathLike` in a follow-up if you want the writer to accept `Path` too.

Verified: `pytest tests/preprocessing/input_output/sdf.py` → 11 passed (added `test_mol_from_sdf_accepts_path_object`); `ruff check` + `ruff format` clean.

### DIFF
--- a/skfp/preprocessing/input_output/sdf.py
+++ b/skfp/preprocessing/input_output/sdf.py
@@ -2,6 +2,7 @@ import os.path
 import warnings
 from collections.abc import Sequence
 from numbers import Integral
+from pathlib import Path
 
 from joblib import effective_n_jobs
 from rdkit.Chem import Mol, SDMolSupplier, SDWriter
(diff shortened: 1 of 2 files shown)


j-adamczyk (approved):

j-adamczyk:
Please merge the latest master commit, I needed to bump up dependencies. With that, CI should pass

rayair250-droid:
Heads-up on the red CI: the failing step is `uv audit`, which now reports 3 recently-disclosed advisories in dev dependencies (`jupyter-server` 2.19.0 and `jupyterlab` 4.5.8) — unrelated to this PR. `mypy`, `ruff check`, `ruff format` and the SDF tests all pass. Any current PR will hit the same audit failure until those are bumped (jupyter-server >= 2.20.0, jupyterlab >= 4.5.9).

PR #632, opened 2026-07-08
Add topological index

j-adamczyk:
## Changes

Topological index for Mordred.


FloudMe77 (approved):

skfp/fingerprints/_new_mordred/descriptors/topological_index.py
```diff
@@ -0,0 +1,45 @@
...
+    petitjean_index = (
+        np.nan
+        if graph_diameter == 0
+        else (graph_diameter - graph_radius) / graph_diameter
+    )
+
+    values = np.asarray(
+        [
+            int(graph_diameter),
+            int(graph_radius),
```
OlaszPL: Do we need casting them to integers? Or that's used instead of math.floor() ?

j-adamczyk: Actually here it's always an integer, the cast is not necessary, good catch

PR #633, opened 2026-07-09
Add Mordred topological charge

j-adamczyk:
## Changes

Adds topological charge descriptors for Mordred. ~10x faster than original.


FloudMe77 (approved):

skfp/fingerprints/_new_mordred/descriptors/topological_charge.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _charge_term_matrix(adj_matrix: np.ndarray, dist_matrix: np.ndarray) -> np.ndarray:
+    """
+    Charge term matrix ``M - M.T``, with ``M = A . D``, where ``A`` is the
+    adjacency matrix and ``D`` the inverse-square topological distance matrix
+    (zero on the diagonal).
+    """
+    inv_sq_dist = dist_matrix.astype(np.float64, copy=True)
+    nonzero = inv_sq_dist != 0
+    inv_sq_dist[nonzero] = inv_sq_dist[nonzero] ** -2
```
OlaszPL, line 76: Do we really need f64 here?

OlaszPL (approved):

skfp/fingerprints/_new_mordred/descriptors/topological_charge.py
j-adamczyk: Maybe not necessarily, but we are inverting values and squaring that, so I thought more precision would be useful just in case

OlaszPL: Ok, makes sense

PR #657, opened 2026-07-23
Add element_counts descriptor (closes #551)

### LINKED ISSUE #551: Descriptor with element counts
We have a function to calculate the number of occurrences of a given element in https://github.com/MLCIL/scikit-fingerprints/blob/master/skfp/descriptors/constitutional.py. It would be useful to have a function to get a vector of counts of atoms from all elements supported by RDKit, for analysis or fingerprint features.

rayair250-droid:
Closes #551.

## Summary

Adds `element_counts(mol)` to `skfp.descriptors.constitutional`, returning a vector of atom counts across all elements of the RDKit periodic table — useful as fingerprint features or for composition analysis, as requested in #551.

## Details

- Returns a length-118 `np.ndarray` indexed by atomic number: position `i` holds the count of the element with atomic number `i + 1` (index 0 = hydrogen, index 5 = carbon, …).
- Hydrogens are counted in total (explicit + implicit), **consistently with the existing `element_atom_count`** — the H bucket is filled by delegating to `element_atom_count(mol, 1)`, so the two functions never disagree.
- Exported from `skfp.descriptors`, added to `docs/modules/descriptors.rst`, and covered by parametrized tests (per-element counts, consistency with `element_atom_count`, absent-element zeros and total sum).

```python
>>> from rdkit.Chem import MolFromSmiles
>>> from skfp.descriptors import element_counts
>>> counts = element_counts(MolFromSmiles("CCO"))  # ethanol
>>> int(counts[5]), int(counts[7]), int(counts[0])  # C, O, H
(2, 1, 6)
```

## Notes

- Doctests use `int(counts[i])` because NumPy 2.x scalars repr as `np.int64(6)`; wrapping keeps the doctest output stable across NumPy versions.
- Happy to switch the return type to a dict keyed by symbol/atomic number instead of a fixed-length vector if you prefer that API — the issue said "vector", so I went literal.

### DIFF
--- a/docs/modules/descriptors.rst
+++ b/docs/modules/descriptors.rst
@@ -27,6 +27,7 @@ Constitutional descriptors (based on atomic composition):
     average_molecular_weight
     bond_count
     element_atom_count
+    element_counts
     heavy_atom_count
     molecular_weight
     number_of_rings
--- a/skfp/descriptors/__init__.py
+++ b/skfp/descriptors/__init__.py
@@ -3,6 +3,7 @@ from .constitutional import (
     average_molecular_weight,
     bond_count,
     element_atom_count,
+    element_counts,
     heavy_atom_count,
     molecular_weight,
     number_of_rings,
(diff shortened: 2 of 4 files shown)


j-adamczyk:
Please refrain from opening new PRs before addressing comments in the previous one.

PR #662, opened 2026-08-06
A few fixes

### LINKED ISSUE #660: Reading from SDF file should create `PropertyMol` rather than `Mol`
By default, RDKit creates `Mol` objects, which do not preserve properties when pickling. This is problematic, since 3D fingerprints with parallelization via Joblib use that. For example, when reading molecules with `MolFromSDFTransformer`, this won't work. It should create `PropertyMol` objects instead. This stems from relying on the `conf_id` property.

j-adamczyk:
## Changes

Fixes https://github.com/MLCIL/scikit-fingerprints/issues/660 and https://github.com/MLCIL/scikit-fingerprints/issues/659, and also a few minor things.

### DIFF
--- a/tests/filters/rule_of_three.py
+++ b/tests/filters/rule_of_three.py
@@ -28,7 +28,7 @@ def smiles_passing_one_violation_basic_rule_of_three() -> list[str]:
     return [
         "c1ccc(CCNc2ccc3nncn3n2)cc1",
         "c1ccc(CCNc2ncnc3nc[nH]c23)cc1",
-        "c1ccc(Cn2nnc3cnccc32)cc1",
+        "CN(C)CCOC(C1=CC=CC=C1)C2=CC=CC=C2",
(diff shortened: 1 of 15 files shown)


tests/filters/rule_of_three.py
```diff
@@ -92,6 +92,7 @@ def test_mols_passing_with_violation_basic_rule_of_three(
     smiles_filtered = mol_filter.transform(
         smiles_passing_one_violation_basic_rule_of_three
     )
+    print(smiles_filtered)
```
mjste: Print in tests?

mjste (dismissed):
A bit long

tests/filters/rule_of_three.py
j-adamczyk: Fixed

my-alaska (approved):

PR #664, opened 2026-08-08
Various fixes

j-adamczyk:
## Changes

A bunch of fixes:
- many typos
- `tqdm` parameter parsing
- better shape checking in MaxMinClustering
- added type cast in distances, since for dense fingerprints (e.g. RDKit) default uint8 can easily result in overflow in bulk computation
- changed Spearman correlation metric to match SciPy for identical vectors
- conformer generation had slightly different results in sequential vs parallel version, since it used double but RDKit pickles always as float; casting to float is harmless (precision is never this high anyway) and gives exactly the same results


skfp/applicability_domain/knn.py
```diff
@@ -158,11 +158,9 @@ def fit(  # noqa: D102
...
         self.X_train_ = X
 
-        # k+1, since we need to exclude each point from being its own neighbor
-        if self.agg == "min":
-            self._k_used = min(len(X), 2)
-        else:
-            self._k_used = min(len(X), self.k + 1)
+        # number of neighbors aggregated when scoring a sample; for "min" only the
+        # single nearest neighbor matters, regardless of k
+        self._k_used = 1 if self.agg == "min" else self.k
```
mjste, line 164: Add assertion - if self.agg is min then self.k must not be specified/be default

skfp/applicability_domain/topkat.py
```diff
@@ -92,21 +92,21 @@ def fit(  # noqa: D102
         self.X_min_ = X.min(axis=0)
         self.X_max_ = X.max(axis=0)
         self.range_ = self.X_max_ - self.X_min_
-        self.num_points = X.shape[0]
-        self.num_dims = X.shape[1]
+        self.num_points_ = X.shape[0]
+        self.num_dims_ = X.shape[1]
```
mjste: What's with the name?

skfp/applicability_domain/topkat.py
```diff
@@ -92,21 +92,21 @@ def fit(  # noqa: D102
...
-        # Augment with bias (1) so the subsequent rotation (PCA) includes the intercept term.
+        # augment with bias (1) so the subsequent rotation (PCA) includes the intercept
         S_bias = np.c_[np.ones(S.shape[0]), S]
 
         cov_matrix = S_bias.T @ S_bias
         eigvals, eigvecs = np.linalg.eigh(cov_matrix)
-        self.eigen_val = np.real(eigvals)
-        self.eigen_vec = np.real(eigvecs)
+        self.eigen_val_ = np.real(eigvals)
+        self.eigen_vec_ = np.real(eigvecs)
```
mjste, line 112: ?

skfp/descriptors/constitutional.py
```diff
@@ -108,7 +108,13 @@ def element_atom_count(mol: Mol, atom_id: int | str) -> int:
...
     """
     if atom_id in (1, "H"):
-        return sum(atom.GetTotalNumHs() for atom in mol.GetAtoms())
+        # includeNeighbors also counts hydrogens present as explicit atoms in the
+        # graph, e.g. after AddHs(), which GetTotalNumHs() alone would miss
+        return sum(
+            atom.GetTotalNumHs(includeNeighbors=True)
+            for atom in mol.GetAtoms()
+            if atom.GetAtomicNum() != 1
+        )
```
mjste: What about explicit hydrogens? Is it counted now?

skfp/metrics/spearman.py
```diff
@@ -21,7 +21,6 @@ def spearman_correlation(
     *,
     alternative: str = "two-sided",
     return_p_value: bool = False,
-    equal_values_result: float = 1.0,
```
mjste, line 24: Removing arguments is not backwards compatible. Just pointng it would be better to release at least as a minor version, not patch.

skfp/utils/parallel.py
```diff
@@ -31,6 +32,39 @@ def print_progress(self) -> None:  # noqa: D102
...
+
+    Returns
+    -------
+    tqdm_settings : dict
+        Keyword arguments for ``tqdm()``.
+    """
+    if isinstance(verbose, (bool, int)):
+        return {"total": total, "disable": not verbose}
+    elif isinstance(verbose, dict):
+        tqdm_settings = verbose.copy()
```
mjste, line 58: Why copy?

tests/applicability_domain/utils.py
```diff
@@ -1,52 +1,113 @@
...
 def get_data_outside_ad(
-    n_train: int = 1000, n_test: int = 100, binarize: bool = False
+    n_train: int = 1000, n_test: int = 100, data_type: str = "continuous"
 ) -> tuple[np.ndarray, np.ndarray]:
-    n_features = 256 if binarize else 2
+    """
+    Training data as a wide blob, test data far away from it, i.e. firmly outside
+    the applicability domain.
+    """
+    n_features = 2 if data_type == "continuous" else 256
```
mjste, line 48: Is there data type list written somewhere? It would be nice to have it as an enum or at least static string

skfp/applicability_domain/knn.py
j-adamczyk: Other values of `k` aren't wrong per se, just not used. Maybe I'll specify this more explicitly in the docstring?

skfp/applicability_domain/topkat.py
j-adamczyk: Fixed

skfp/descriptors/constitutional.py
j-adamczyk: Yes

skfp/metrics/spearman.py
j-adamczyk: Ok, I can revert this. But changing the default to `np.nan` makes sense IMO, to follow SciPy

skfp/utils/parallel.py
j-adamczyk: Because we are mutating the user-provided dictionary

tests/applicability_domain/utils.py
j-adamczyk: Good idea, I've added it

skfp/metrics/spearman.py
j-adamczyk: After consideration I changed this back, no major reason for breaking change unless users ask for it

skfp/applicability_domain/topkat.py
j-adamczyk: This follows scikit-learn naming convention. Previous naming was not

my-alaska (dismissed):

skfp/applicability_domain/knn.py
mjste: It was meant as an explicit signal to user that other configurations are ineffective/invalid

j-adamczyk: I added a warning

mjste (approved):

PR #661, opened 2026-08-02
Mordred optimization

j-adamczyk:
## Changes

Heavy optimizations for Mordred. Goes from ~2x faster than original to ~10x faster.

## Checklist before requesting a review
- [ ] Docstrings added/updated in public functions and classes
- [ ] Tests added, reasonable test coverage (at least ~90%, `make test-coverage`)
- [ ] Sphinx docs added/updated and render properly (`make docs` and see `docs/_build/index.html`)

PR #665, opened 2026-08-09
Mordred: resolve feature names per module instead of per call

j-adamczyk:
Every descriptor returned its values together with its feature names, and
the calculator looked up an index for each name on every molecule. The names
never change, so this resolves them once per module at import time and lets
the descriptors return values alone.

- feature_names.py keeps FEATURE_NAMES_2D and FEATURE_NAMES_3D, with
  ALL_FEATURE_NAMES derived from them, instead of spelling out a third list
  that duplicated both (-1600 lines)
- calculator.py declares MODULES_2D/MODULES_3D and precomputes the output
  positions of every module in _OUTPUT_IDXS_2D/_OUTPUT_IDXS_3D
- descriptors return np.ndarray, and tests read the names from the module

Two ordering changes follow from deriving ALL_FEATURE_NAMES:

- the 3D output now appends the conformer features after the 2D ones rather
  than interleaving them, so a 2D feature sits at the same position whether
  or not 3D is used; the 3D tests rearrange mordred's columns to match
- the 24 BCUT columns move within the 2D block (all NaN, as BCUT is not
  wired into the calculator yet)

Values are otherwise unchanged: on 300 molecules sampled from MoleculeNet,
TDC and MoleculeACE, the 2D output matches the previous implementation
exactly once columns are aligned by name.

---

Splits up #661 into reviewable pieces. Based on `mordred` - review and merge in order.

<details>
<summary>The stack</summary>

1. #665 - feature names resolved per module  **<- this PR**
2. #666 - AtomicProperties
5. #669 - RingSets
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
13. #677 - walk counts
</details>


skfp/fingerprints/_new_mordred/utils/feature_names.py
```diff
@@ -2479,30 +648,30 @@
...
     "BCUT_ionization_potential_largest_eigval",
-    "BCUT_valence_electrons_smallest_eigval",
-    "BCUT_valence_electrons_largest_eigval",
-    "BCUT_sigma_electrons_smallest_eigval",
-    "BCUT_sigma_electrons_largest_eigval",
-    "BCUT_intrinsic_state_smallest_eigval",
-    "BCUT_intrinsic_state_largest_eigval",
-    "BCUT_gasteiger_charge_smallest_eigval",
-    "BCUT_gasteiger_charge_largest_eigval",
+    "BCUT_ionization_potential_smallest_eigval",
```
OlaszPL, line 666: This change seems strange but I don't know why

tests/fingerprints/new_mordred_fp.py
```diff
@@ -14,6 +27,20 @@ def _logee_mask(feature_names: np.ndarray) -> np.ndarray:
...
+    Rearrange mordred's columns into this fingerprint's 3D layout, which appends
+    the conformer descriptors after the 2D ones instead of interleaving them.
+
+    Within each of the two blocks the layouts agree position by position, so the
+    rearrangement needs no names -- which matters for the descriptor families that
+    this implementation exposes under different names.
+    """
+    names_2d = MordredFingerprint().get_feature_names_out()
+    is_2d = np.isin(feature_names_old, names_2d)
+    return np.hstack([X_old[:, is_2d], X_old[:, ~is_2d]])
```
OlaszPL: From what I understand it gets MordredFingerprint instance names and then names from always 2d instance and sorts by that?

skfp/fingerprints/_new_mordred/calculator.py
```diff
@@ -56,8 +58,83 @@
...
+    ``FEATURE_NAMES_3D``; the rest declare a single ``FEATURE_NAMES``.
+    """
+    split_names = getattr(
+        module, "FEATURE_NAMES_3D" if use_3D else "FEATURE_NAMES_2D", None
+    )
+    return split_names if split_names is not None else module.FEATURE_NAMES
+
+
+_OUTPUT_IDXS_2D = _output_idxs(MODULES_2D, use_3D=False)
+_OUTPUT_IDXS_3D = _output_idxs(MODULES_3D, use_3D=True)
```
OlaszPL, line 139: Maybe extract that into a new file in `utils/`? It takes more than half of this file and is only responsible for feature names handling

OlaszPL (changes requested):

skfp/fingerprints/_new_mordred/calculator.py
j-adamczyk: I considered this, but note that this requires a list of actual modules that compute descriptors. I would prefer not to make that massive import in another file, since that drastically increases the chance for problems.

tests/fingerprints/new_mordred_fp.py
```diff
@@ -1,44 +1,916 @@
...
+    "SpMAD_Dzionization_potential": "SpMAD_Dzi",
+    "LogEE_Dzionization_potential": "LogEE_Dzi",
+    "SM1_Dzionization_potential": "SM1_Dzi",
+    "VE1_Dzionization_potential": "VE1_Dzi",
+    "VE2_Dzionization_potential": "VE2_Dzi",
+    "VE3_Dzionization_potential": "VE3_Dzi",
+    "VR1_Dzionization_potential": "VR1_Dzi",
+    "VR2_Dzionization_potential": "VR2_Dzi",
+    "VR3_Dzionization_potential": "VR3_Dzi",
+}
```
OlaszPL: Could we extract that?

tests/fingerprints/new_mordred_fp.py
```diff
@@ -1,44 +1,916 @@
...
-
-    Within each of the two blocks the layouts agree position by position, so the
-    rearrangement needs no names -- which matters for the descriptor families that
-    this implementation exposes under different names.
-    """
-    names_2d = MordredFingerprint().get_feature_names_out()
-    is_2d = np.isin(feature_names_old, names_2d)
-    return np.hstack([X_old[:, is_2d], X_old[:, ~is_2d]])
+def _to_mordred_name(name: str) -> str:
+    """Name that mordred gives the feature which this implementation calls ``name``."""
```
OlaszPL: Fix newlines

tests/fingerprints/new_mordred_fp.py
```diff
@@ -200,27 +1067,44 @@ def test_new_mordred_3D_feature_names():
...
+    )
+    return df_new, df_old[df_new.columns]
+
+
+def _get_features_to_compare(df_new: pd.DataFrame, df_old: pd.DataFrame) -> list[str]:
+    not_computed = (df_new.isna() | df_old.isna()).all()
+    return [
+        name
+        for name in df_new.columns
+        if not not_computed[name] and "LogEE" not in name
```
OlaszPL, line 189: Is it possible to xfail LogEE instead of skipping?

tests/fingerprints/new_mordred_fp.py
j-adamczyk: done

tests/fingerprints/new_mordred_fp.py
j-adamczyk: Not really, because we are using `assert_allclose`. But this is why we have a separate test `test_new_mordred_logee_diverges`

tests/fingerprints/new_mordred_fp.py
j-adamczyk: Sure

OlaszPL (approved):
Awesome refactor!

pkacprzak5 (approved):

FloudMe77 (approved):

PR #666, opened 2026-08-09
Mordred: read per-atom properties into arrays once per molecule

j-adamczyk:
The descriptors each walked the molecule themselves, calling RDKit once per
atom or bond for the same quantities - atomic numbers, degrees, hydrogen
counts, valence electrons, Gasteiger charges. AtomicProperties reads them
into NumPy arrays once and hands the same arrays to every descriptor.

- AtomicProperties.from_mol() collects the per-atom and per-bond arrays, with
  the derived ones (valence electrons, intrinsic state, neighbor counts)
  computed lazily and cached
- with_hydrogens_added() builds the hydrogen-explicit properties from the
  heavy-atom ones instead of re-reading the molecule
- PeriodicTable.lookup() indexes a padded array with an atomic number vector,
  for whole-molecule lookups
- the calculator builds props_regular and props_hydrogens and passes them on

Four descriptors move to it here as the first consumers: aromatic,
fragment_complexity, polarizability and vertex_adjacency_info. The rest
follow one PR at a time; PROPERTY_FUNCS and PERIOD stay for now, since the
descriptors that have not moved yet still read a property at a time.

Output is unchanged: identical values on 300 molecules sampled from
MoleculeNet, TDC and MoleculeACE.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-01-feature-names` - review and merge in order.

<details>
<summary>The stack</summary>

1. #665 - feature names resolved per module
2. #666 - AtomicProperties  **<- this PR**
3. #667 - hydrogen-explicit matrices derived
4. #668 - spectral attributes over a stack
5. #669 - RingSets
6. #670 - subgraph enumeration (chi, path counts)
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
11. #675 - E-state indices shared with VSA
12. #676 - BCUT (not computed before)
13. #677 - walk counts
14. #678 - conformer descriptors
15. #679 - atom, bond and constitutional counts
16. #680 - topological charge and Zagreb
17. #681 - ABC and molecular distance edge
</details>

PR #667, opened 2026-08-09
Mordred: derive the hydrogen-explicit matrices instead of recomputing them

j-adamczyk:
The distance matrix of the hydrogen-explicit molecule was computed from
scratch, an all-pairs shortest path over a molecule several times the size of
the heavy-atom one. Added hydrogens are terminal atoms, so their distances
follow from the heavy-atom distances: one bond to the atom a hydrogen hangs
off, and one at each end between two hydrogens.

- DistanceMatrix now wraps a matrix, built either by from_mol() or by
  with_hydrogens_added() from the heavy-atom distances
- eccentricities, radius and diameter are computed once in the constructor
  rather than on every property access
- RDKit's distance matrix cache is keyed per flag combination, so that the
  bond-order and atom-weight variants no longer evict each other
- the calculator derives the hydrogen-explicit molecule with AddHs(mol_regular)
  and its properties with AtomicProperties.with_hydrogens_added()

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-02-atomic-properties` - review and merge in order.

<details>
<summary>The stack</summary>

2. #666 - AtomicProperties
3. #667 - hydrogen-explicit matrices derived  **<- this PR**
5. #669 - RingSets
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
13. #677 - walk counts
</details>


tests/fingerprints/_new_mordred/autocorrelation.py
```diff
@@ -34,7 +34,7 @@ def computed_values():
     for name, smiles in _REFERENCE["SMILES"].items():
         # Mordred uses explicit-hydrogen molecules
         mol = AddHs(MolFromSmiles(smiles))
-        values = calc(mol, DistanceMatrix(mol))
+        values = calc(mol, DistanceMatrix.from_mol(mol))
```
OlaszPL, line 37: I know that's not the best place, but wouldn't it be nice to make such a comparison test?
a = DistanceMatrix.from_mol(mol_regular)
b = DistanceMatrix.with_hydrogens_added(mol_hydrogens, a)
c = DistanceMatrix.from_mol(mol_hydrogens)
assert b == c ?

If so, it'd be nice to make a similar test for AtomicProperties too.

skfp/fingerprints/_new_mordred/utils/graph_matrix.py
```diff
@@ -12,72 +12,137 @@
...
-            self._orders.append(self._orders[-1].dot(self._base))
-        return self._orders[n - 1]
-
-    @cached_property
-    def degree(self) -> np.ndarray:
-        """Number of edges incident to each vertex (atom).
-
-        By default ``use_bond_orders=False``, so bond orders are ignored and each
-        bond counts as one edge. In that case, atom degree equals the atom
-        valence, i.e. the number of bonds each atom forms.
```
OlaszPL, line 61: I believe this comment was important, because mordredcommunity calls degree as valence in this place - worth mentioning to keep consistency

j-adamczyk: Reverted

tests/fingerprints/_new_mordred/autocorrelation.py
j-adamczyk: Done, and actually found a bug that I also fixed

OlaszPL (approved):

FloudMe77 (approved):

PR #668, opened 2026-08-09
Mordred: compute the spectral attributes of a stack of matrices at once

j-adamczyk:
The thirteen spectral attributes were thirteen cached properties, each
reducing over the eigenvalues on its own, and every matrix was decomposed by
its own MatrixAttributes instance. They are now computed together by
spectral_attributes(), which takes a stack of matrices and returns an
(n_matrices, 13) array.

This matters most for the Barysz descriptors, which ask for the attributes of
eight matrices of the same molecule: the reductions run over arrays of a few
dozen numbers, where NumPy's dispatch costs more than the arithmetic, so doing
them once for the whole stack is what saves the time. MatrixAttributes stays
as the single-matrix front end, a stack of one.

- adjacency_matrix reuses the eigendecomposition the calculator now computes
  once, instead of decomposing the same matrix again
- adjacency_matrix and distance_matrix take AtomicProperties; barysz_matrix
  and detour_matrix pass theirs through to MatrixAttributes, their own bodies
  are untouched here

Values are unchanged except for VR1_Dt, VR2_Dt and VR3_Dt, which differ by up
to 4e-7 relative on 300 molecules sampled from MoleculeNet, TDC and
MoleculeACE - float32 rounding from summing the Randic-like terms in a
different order. Everything else is identical.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-03-graph-matrices` - review and merge in order.

<details>
<summary>The stack</summary>

1. #665 - feature names resolved per module
2. #666 - AtomicProperties
3. #667 - hydrogen-explicit matrices derived
4. #668 - spectral attributes over a stack  **<- this PR**
5. #669 - RingSets
6. #670 - subgraph enumeration (chi, path counts)
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
11. #675 - E-state indices shared with VSA
12. #676 - BCUT (not computed before)
13. #677 - walk counts
14. #678 - conformer descriptors
15. #679 - atom, bond and constitutional counts
16. #680 - topological charge and Zagreb
17. #681 - ABC and molecular distance edge
</details>

PR #669, opened 2026-08-09
Mordred: perceive the rings of a molecule once, in RingSets

j-adamczyk:
Ring perception was repeated per descriptor: ring_count called GetSymmSSSR
and then walked every ring to check aromaticity and heteroatoms, and
vdw_volume_abc did the same work again through the private helpers of
ring_count. RingSets does it once per molecule, caching the simple and the
fused ring atom sets together with the size, aromaticity and heteroatom flags
of each ring.

- ring counts are read off boolean masks over the cached ring properties,
  rather than recomputing the predicates per ring size bucket
- vdw_volume_abc takes the shared RingSets and the hydrogen-explicit
  properties, and its Bondi volumes are a table indexed by atomic number
  instead of a dict keyed by element symbol
- the calculator builds rings_regular next to props_regular

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-04-matrix-attributes` - review and merge in order.

<details>
<summary>The stack</summary>

2. #666 - AtomicProperties
5. #669 - RingSets  **<- this PR**
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
13. #677 - walk counts
</details>


tests/fingerprints/_new_mordred/vdw_volume_abc.py
```diff
@@ -25,5 +27,6 @@ def test_vdw_volume_abc_values(smiles, expected):
     mol_hydrogens = preprocess_mol(mol, explicit_hydrogens=True)
     mol_regular = preprocess_mol(mol)
```
OlaszPL:
```py
mol_regular = preprocess_mol(mol)
mol_hydrogens = AddHs(mol_regular)
```
Will be faster

tests/fingerprints/_new_mordred/vdw_volume_abc.py
```diff
@@ -25,5 +27,6 @@ def test_vdw_volume_abc_values(smiles, expected):
...
 
-    values = vdw_volume_abc.calc(mol_regular, mol_hydrogens)
+    rings = RingSets(mol_regular, AtomicProperties.from_mol(mol_regular))
+    values = vdw_volume_abc.calc(rings, AtomicProperties.from_mol(mol_hydrogens))
```
OlaszPL:
```py
props_regular = AtomicProperties.from_mol(mol_regular)
props_hydrogens = AtomicProperties.with_hydrogens_added(props_regular, mol_hydrogens)
```
Will be faster too

skfp/fingerprints/_new_mordred/descriptors/vdw_volume_abc.py
```diff
@@ -54,24 +57,23 @@ def calc(mol_regular: Mol, mol_hydrogens: Mol) -> np.ndarray:
...
         return np.asarray([np.nan], dtype=np.float32)
 
-    n_bonds = mol_hydrogens.GetNumBonds()
+    n_bonds = props_hydrogens.num_bonds
 
     # reuse ring detection from the ring count descriptor: simple (non-fused)
     # aromatic and non-aromatic rings
-    rings = _ring_properties(mol_regular, _ring_atom_sets(mol_regular))
+    rings = rings_regular.simple_rings
     n_aromatic_rings = sum(1 for ring in rings if ring.is_aromatic)
```
OlaszPL: Isn't there a more "numpy way" to calculate this?

skfp/fingerprints/_new_mordred/descriptors/ring_count.py
```diff
@@ -35,6 +37,41 @@ def size(self) -> int:
...
+        self.fused_rings = self._ring_properties(
+            _fused_ring_atom_sets(self.simple_ring_atom_sets)
+        )
+
+    def _ring_properties(self, ring_atom_sets: list[set[int]]) -> list[RingProperties]:
+        """
+        Cache ring size, aromaticity, and heteroatom presence once per ring.
+        """
+        is_aromatic = self._props.is_aromatic
+        is_hetero = self._props.atomic_nums != 6
```
OlaszPL: Shouldn't that be moved into AtomicProperties? I believe it was used somewhere in previous PRs.

skfp/fingerprints/_new_mordred/descriptors/vdw_volume_abc.py
j-adamczyk: This can be slightly simplified, but changing to NumPy arrays would be marginally beneficial, since those are very short lists

skfp/fingerprints/_new_mordred/descriptors/ring_count.py
j-adamczyk: Good idea, it's used in 2 places, but will be slightly cleaner

OlaszPL (approved):

skfp/fingerprints/_new_mordred/calculator.py
FloudMe77: GetSymmSSSR runs a third time here, but doesn't rings_regular.num_rings already hold this? SSSR count is the circuit rank (E − V + C), invariant under kekulization, so it should match (checked on 400 BBBP mols, 0 mismatches). Could we just pass rings_regular.num_rings and drop this call?

skfp/fingerprints/_new_mordred/descriptors/ring_count.py
```diff
@@ -35,6 +37,41 @@ def size(self) -> int:
...
+    def _ring_properties(self, ring_atom_sets: list[set[int]]) -> list[RingProperties]:
+        """
+        Cache ring size, aromaticity, and heteroatom presence once per ring.
+        """
+        is_aromatic = self._props.is_aromatic
+        is_hetero = self._props.is_hetero
+        return [
+            RingProperties(
+                ring,
+                bool(is_aromatic[list(ring)].all()),
```
FloudMe77: I will extract `list(ring)` into an external variable to avoid materializing the list twice.

FloudMe77 (approved):

skfp/fingerprints/_new_mordred/descriptors/ring_count.py
j-adamczyk: Ok, it's faster

skfp/fingerprints/_new_mordred/calculator.py
j-adamczyk: SSSR is not necessarily a circuit rank. It can return slightly more rings on highly symmetric molecules, see e.g. cubane for an illustrative edge case. It's not very expensive overall, but we can revisit how many times we call SSSR later.

PR #670, opened 2026-08-09
Mordred: enumerate the connected subgraphs once, with array operations

j-adamczyk:
The chi and path count descriptors each enumerated the connected subgraphs of
the molecule themselves, in Python, one subgraph at a time - together the
largest cost in the whole calculator, around 29% of its Python time. Subgraphs
enumerates them once per molecule: subgraphs of each order are grown from
those of the previous order by adding one adjacent bond, entirely with array
operations over flat index arrays.

- utils/subgraphs.py enumerates up to order 7, the largest the chi descriptors
  ask for, and classifies each subgraph as chain, path, path cluster or
  cluster
- the atom index sets of a type are gathered only when a descriptor asks for
  them, since chi wants different types at different orders
- utils/ragged.py holds the two helpers for rows of unequal length that this
  bookkeeping needs
- chi and path_count read the shared Subgraphs and the shared properties

Computing all 2D descriptors for 400 molecules drops from 19.5s to 8.5s.

Values are unchanged except for 54 chi columns (Xc, Xch, Xp, Xpc, AXp), which
differ by up to 3.4e-6 relative on 300 molecules sampled from MoleculeNet,
TDC and MoleculeACE - float32 rounding from summing the per-subgraph terms in
a different order.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-05-ring-sets` - review and merge in order.

<details>
<summary>The stack</summary>

1. #665 - feature names resolved per module
2. #666 - AtomicProperties
3. #667 - hydrogen-explicit matrices derived
4. #668 - spectral attributes over a stack
5. #669 - RingSets
6. #670 - subgraph enumeration (chi, path counts)  **<- this PR**
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
11. #675 - E-state indices shared with VSA
12. #676 - BCUT (not computed before)
13. #677 - walk counts
14. #678 - conformer descriptors
15. #679 - atom, bond and constitutional counts
16. #680 - topological charge and Zagreb
17. #681 - ABC and molecular distance edge
</details>

PR #671, opened 2026-08-09
Mordred: build the detour matrix from the distance matrix and its rings

j-adamczyk:
The detour matrix was built by splitting the molecular graph into biconnected
blocks with networkx, solving the longest simple paths inside each block by
brute-force DFS, and stitching the blocks back together through their
articulation nodes. The stitching, the graph conversion and the networkx
dependency all go away here.

Outside the rings there is only one simple path between two atoms, so the
detour distance is the ordinary distance; a path can differ from the shortest
one only where it crosses a ring system. Each ring system is added on top of
the distance matrix as a correction - how much longer the longest way through
it is than the shortest one - read off at the atoms where a path enters and
leaves the system. A simple path cannot leave a ring system and come back, so
the corrections are independent of one another. Plain rings need no search at
all: the two ways round between two atoms make up the whole ring.

The ring systems come from merging the perceived rings that share at least two
atoms, which is not always the whole story: RDKit's rings do not cover every
cycle of a cage-like molecule, and two rings sharing a single atom can still
belong to one cage, where a path leaves through one shared atom and returns
through another. Both leave a cycle spanning more than one ring system, which
_are_ring_systems_complete() detects by comparing the cycle rank of the
molecule against the ranks of its ring systems, in one pass over the bonds.
When they do not match, the systems are taken from the components left after
removing the bridges, which is exact and costs an extra DFS.

Verified against the previous implementation on every molecule of MoleculeNet,
TDC and MoleculeACE - 591,388 connected molecules, of which 19,735 disconnected
ones return NaN under both. The matrices are identical everywhere. Without the
completeness check one molecule differs (an HIV di-iron cage, DetourIndex 5441
against the correct 9633), and the check flags exactly that one. Fuzzing over
23k random polycyclic skeletons against brute-force longest simple paths finds
48 more such cages, all of them flagged, with no false alarms.

The 14 detour features differ by up to 2e-7 relative, from the corrections
being accumulated in the float64 distance matrix. The check itself costs about
7% of the descriptor, and its fallback fired on 1 molecule in 591,388.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-06-subgraphs` - review and merge in order.

<details>
<summary>The stack</summary>

2. #666 - AtomicProperties
5. #669 - RingSets
7. #671 - detour matrix  **<- this PR**
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices
13. #677 - walk counts
</details>


skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -75,60 +86,308 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+
+    A simple path can never leave a ring system and come back, since that would
+    mean visiting the atom it left through twice, which is what makes the
+    corrections of the ring systems independent of one another.
     """
-    n = mol.GetNumAtoms()
+    ring_systems = _get_ring_systems(ring_atom_sets)
 
-    if n == 1:
-        return np.array([[0]], dtype=np.float32)
```
OlaszPL, line 90: Why not?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -75,60 +86,308 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+                rest.append(system)
+        systems = [*rest, merged]
+
+    # merging two systems can bring a third within reach, so repeat until stable
+    while True:
+        for first, second in itertools.combinations(systems, 2):
+            if len(first & second) >= 2:
+                systems.remove(second)
+                first |= second
+                break
```
OlaszPL, line 149: Is this guaranteed to work? We check every combination and the first one which fires the inner if breaks the loop. What's the confidence that the next combination wouldn't fire this if again? Simply put, won't that `break` make us omit some cases?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -75,60 +86,308 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+    no cycle spans more than one of them, and that is not guaranteed by merging
+    the perceived rings. Cage-like molecules break it in two ways: RDKit's rings
+    do not always cover every cycle of such a molecule, and two rings sharing a
+    single atom can still be part of one cage, where a path leaving through one
+    of the shared atoms comes back through another - unlike a genuine spiro atom.
+
+    Both show up as cycles unaccounted for, which the cycle rank counts. The rank
+    of the whole molecule (``bonds - atoms + 1``, for a connected one) splits over
+    the parts it cannot be traversed between, so it equals the ranks of the ring
+    systems summed, and any cycle escaping them makes that sum come out lower.
```
OlaszPL:
```suggestion
    Both show up as cycles unaccounted for, which is what comparing two cycle
    ranks catches: the one of the whole molecule against the ones of the ring
    systems summed. The rank counts not how many cycles something has but how
    many of them are independent, the number of bonds it carries over a spanning
    tree, each of them closing one cycle.

    The rank of the molecule (``bonds - atoms + 1``, for a connected one) splits
    over the parts it cannot be traversed between, so the two come out equal when
    the systems hold every cycle. Sharing no bonds, the systems are always
    independent of one another, so their sum can never come out higher. It comes
    out lower when a cycle has bonds no single system holds, whether it was left
    out of the perceived rings or split between two systems: it counts towards the
    rank of the molecule, but towards none of theirs.
```

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: Fixed

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: Added more comments on this, but yes, it will always work, as it repeats the outer `while`

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
OlaszPL: @j-adamczyk ?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
OlaszPL: I believe you didn't? Diff shows no added comments. Also previous ring system merge seems to be removed? I don't understand?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -1,6 +1,8 @@
 import itertools
 
 import numpy as np
+from scipy.sparse import coo_array
+from scipy.sparse.csgraph import connected_components
```
OlaszPL, line 5: What're these for?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -131,16 +135,7 @@ def _get_ring_systems(ring_atom_sets: list[set[int]]) -> list[set[int]]:
...
-    for ring in ring_atom_sets:
-        merged = set(ring)
-        rest = []
-        for system in systems:
-            if len(system & merged) >= 2:
-                merged |= system
-            else:
-                rest.append(system)
-        systems = [*rest, merged]
+    systems = [set(ring) for ring in ring_atom_sets]
```
OlaszPL: I don't understand this change. It seems that the whole ring system merge part has beed removed? Why?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
OlaszPL: Would you consider applying it?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -178,13 +173,13 @@ def _are_ring_systems_complete(
     rank_sum = 0
     is_inside = np.zeros(props.num_atoms, dtype=bool)
     for ring_system in ring_systems:
-        atoms = np.fromiter(ring_system, dtype=np.intp, count=len(ring_system))
+        atoms = list(ring_system)
```
OlaszPL: Maybe `ring_system_atoms` instead for clarity?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -328,100 +307,72 @@ def _longest_paths_in_ring_system(
...
 
-        result[v] = max(result[v], dist)
+        longest[neighbor] = max(longest[neighbor], distance)
 
-        visited.add(v)
-        _dfs(v, dist, G, visited, result)
-        visited.remove(v)
+        visited.add(neighbor)
+        _walk_every_path(neighbor, distance, neighbors, visited, longest)
+        visited.remove(neighbor)
```
OlaszPL: What's the purpose of this change? That code seemed right before.

OlaszPL: Ok maybe I understand it, but it would be nice to describe it at least in the PR comment because the change was pretty big - and not every decision I understand

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: Should be fixed now

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: Done

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: Graph is represented as COO array now, because this way we can call `connected_components` instead of implementing it

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: It wasn't removed really, but replaced with `while` loop. Previous loop didn't terminate correctly for fused cases

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: done

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
j-adamczyk: This is still a DFS, but should be faster

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -42,11 +51,15 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+        atomic_props_regular,
+        distance_matrix_regular.matrix,
+        rings_regular.simple_ring_atom_sets,
+    )
     attrs = MatrixAttributes(
         detour_matrix,
-        props_regular,
-        hermitian=True,  # as in Mordred's reference implementation
+        atomic_props_regular,
+        hermitian=True,
```
pkacprzak5, line 62: I see the comment about this `hermitian=True` was removed, is there any particular reason? Also as `distance_matrix.calc` uses `hermitian=distance_matrix_regular.hermitian` and here we hardcode it, maybe it is worth adding here that detour is always symmetric, so it is clearer why we can hardcode it here?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -75,154 +88,302 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+
+    A simple path can never leave a ring system and come back, since that would
+    mean visiting the atom it left through twice, which is what makes the
+    corrections of the ring systems independent of one another.
+    """
+    ring_systems = _get_ring_systems(ring_atom_sets)
+
+    # the rings are the ones RDKit perceives, which for cage-like molecules do not
+    # always describe the whole cycle structure, see _are_ring_systems_complete()
+    if not _are_ring_systems_complete(props, ring_systems):
```
pkacprzak5, line 116: Maybe we should add a unit test for this case? A small one, with a known cage SMILES that asserts `_are_ring_systems_complete` is `False` and `DetourIndex` matches the expected value?

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -75,154 +88,302 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+    at a single (spiro) atom cannot: a path through one of them and into the other
+    would have to pass that atom twice.
+    """
+    systems = [set(ring) for ring in ring_atom_sets]
+
+    # merging two systems can bring a third within reach, one sharing a single
+    # atom with each of them and two with the two of them together, so a single
+    # pass is not enough. Every merge breaks out to start the scan over, and the
+    # systems are returned once a whole pass finds no pair left to merge.
+    while True:
```
pkacprzak5, line 144: Isn't this part similar to what happens in `ring_count.py`? In the `_fused_ring_atom_sets`, where we do the union-find?

They use the same merge rule, but different algorithms (Find&Union vs this while True).

Maybe we should extract a shared helper that returns all ring systems including singles?  `_fused_ring_atom_sets` would just later drop them (or we can have one shared helper with a parameter whether we want those singles or not), and we would have one shared helper to prevent mismatches. Also I feel like the Find&Union makes these merges easier to follow compared to this behavior.

skfp/fingerprints/_new_mordred/descriptors/detour_matrix.py
```diff
@@ -75,154 +88,302 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+    the same atom never traverse the system, and their correction is zero.
     """
-    Longest simple path length from a single source ``s`` to every node.
+    within = shortest_distances[atoms[:, np.newaxis], atoms]
+    longest = _longest_paths_in_ring_system(props, atoms, within)
 
-    Returns a mapping ``node -> distance``, with 0 for the source itself.
+    # every atom reaches the ring system through the one of its atoms closest to
+    # it, since all of its paths into the system pass through that atom
+    gates = np.argmin(shortest_distances[:, atoms], axis=1)
```
pkacprzak5, line 311: What happens when we have more than one attom within the same shortest distance? `np.argmin` picks the first one, but is it always safe for the later correction? Can it happen, that choosing a different atom from minimum path would result in a bigger distance?

pkacprzak5 (changes requested):

PR #672, opened 2026-08-09
Mordred: compute the ETA descriptors over property arrays

j-adamczyk:
The extended topochemical atom descriptors walked the kekulized molecule atom
by atom and bond by bond, building per-atom alpha and epsilon values in Python
loops, and asked for two molecules of their own: a kekulized copy and a
kekulized hydrogen-explicit copy, each with its own distance matrix.

Kekulization changes neither the atoms nor the skeleton, so the properties and
the distances of the hydrogen-suppressed molecule apply to the kekulized one
as well. The descriptor now takes those, plus the kekulized bond types as a
plain array, and computes the alpha and epsilon values as array operations.

- the kekulized copy is made from mol_regular rather than re-preprocessed from
  the input, so its atom numbering matches props_regular
- the separate kekulized distance matrix and the extra SSSR call are gone; the
  ring count comes from the shared RingSets
- PERIOD, kept as an alias while this descriptor still read the period one
  atom at a time, is removed

The 17 ETA columns that touch these sums differ by at most 3e-5 absolute
(1.2e-5 relative) on 300 molecules sampled from MoleculeNet, TDC and
MoleculeACE, from float32 accumulation order. The rest are identical.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-07-detour-matrix` - review and merge in order.

<details>
<summary>The stack</summary>

2. #666 - AtomicProperties
5. #669 - RingSets
7. #671 - detour matrix
8. #672 - ETA descriptors  **<- this PR**
9. #673 - autocorrelations
10. #674 - Barysz matrices
13. #677 - walk counts
</details>


skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
pkacprzak5, line 23: Is this used anymore? I do not see it used anywhere, can it be removed?

pkacprzak5, line 190: same about this function, it is not called at all, do we need it, or can it be removed?

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -65,35 +72,37 @@
...
-    )
-    gamma, beta_sigma, beta_non_sigma, beta_delta = _beta_and_gamma(
-        mol_kekulized, atomic_nums, core_counts, epsilons
-    )
+    atomic_nums = props.atomic_nums
+    core_counts, epsilons = _core_counts_and_epsilons(atomic_nums)
+    degrees = props.degrees
+
+    ring_atoms = np.zeros(num_atoms, dtype=bool)
+    ring_atoms[[atom for ring in rings.simple_ring_atom_sets for atom in ring]] = True
```
pkacprzak5, line 102: Can this be moved to `RingSets` maybe? Wouldn't it make code here cleaner, as the `RingSets` keeps all information needed for this either way, so that would simplify the code in this file

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -417,6 +428,121 @@ def _epsilon_values(
...
+    Its carbons fill their remaining bonds with hydrogens, so only how many atoms of
+    either element there are matters, not how they are arranged.
+    """
+    num_carbons = len(degrees)
+    num_hydrogens = num_carbons * _MAX_CARBON_DEGREE - degrees.sum()
+    total = num_carbons * _CARBON_EPSILON + num_hydrogens * _HYDROGEN_EPSILON
+    return float(total / (num_carbons + num_hydrogens))
+
+
 def build_reference_mol(
```
pkacprzak5, line 546: same about this func, i do not see it used anywhere, can it be removed?

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -368,38 +386,31 @@ def _branching_indices(
...
-        atom.GetIdx()
-        for atom in mol_hydrogens.GetAtoms()
-        if atom.GetAtomicNum() != 1 or atom.GetNeighbors()[0].GetAtomicNum() != 6
-    ]
-    eps_5 = eps_hydrogens[keep].mean()
+    is_hydrogen = props_hydrogens.is_hydrogen
+    bonded_to_carbon = np.zeros(props_hydrogens.num_atoms, dtype=bool)
+    begins, ends = props_hydrogens.bond_begin_idxs, props_hydrogens.bond_end_idxs
+    bonded_to_carbon[begins] = atomic_nums[ends] == 6
+    bonded_to_carbon[ends] |= atomic_nums[begins] == 6
```
pkacprzak5, line 412: I feel like those `6` numbers here are not clear enough for me, what do they mean, maybe add some comment what is that 6 and what are we checking here exactly? Although I do not have knowledge about chemistry, so maybe it is clear for someone with it?

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -417,6 +428,121 @@ def _epsilon_values(
     )
 
 
+# the alkane reference replaces every heavy atom with a carbon, which caps the
+# degree it can have, and every bond with a single one
+_MAX_CARBON_DEGREE = 4
+
+# bond order per bond type, for the kekulized bonds read above
+_BOND_ORDER_OF_TYPE = np.full(max(int(t) for t in BondType.values) + 1, np.nan)
```
pkacprzak5, line 436: Isn't `_BOND_ORDER_OF_TYPE` the same table as `_BOND_ORDERS` in `atomic_properties.py`. Can ETA import that instead of keeping a second copy?

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -368,38 +386,31 @@ def _branching_indices(
...
-    keep = [
-        atom.GetIdx()
-        for atom in mol_hydrogens.GetAtoms()
-        if atom.GetAtomicNum() != 1 or atom.GetNeighbors()[0].GetAtomicNum() != 6
-    ]
-    eps_5 = eps_hydrogens[keep].mean()
+    is_hydrogen = props_hydrogens.is_hydrogen
+    bonded_to_carbon = np.zeros(props_hydrogens.num_atoms, dtype=bool)
+    begins, ends = props_hydrogens.bond_begin_idxs, props_hydrogens.bond_end_idxs
+    bonded_to_carbon[begins] = atomic_nums[ends] == 6
```
pkacprzak5, line 411: also why do we have here = and down below |= ? it is not entirely clear to me

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -246,37 +260,41 @@ def _beta_and_gamma(
     return gamma, beta_sigma, beta_non_sigma, beta_delta
 
 
-def _nonsigma_contribute(bond, epsilons: np.ndarray) -> float:
+def _sum_over_bonds(
```
pkacprzak5, line 263: Isn't it the weighted version of `AtomicProperties._count_neighbors`. Maybe we can put it in `AtomicProperties`? `_beta_delta` and `_saturated_mean_epsilon` do the same gather. I'm not sure about whether it would be more readable then though.

skfp/fingerprints/_new_mordred/descriptors/extended_topochemical_atom.py
```diff
@@ -417,6 +428,121 @@ def _epsilon_values(
...
+    )
+    if num_hydrogens is None:
+        return np.nan
+
+    epsilons = _core_counts_and_epsilons(props.atomic_nums)[1]
+    total = epsilons.sum() + num_hydrogens * _HYDROGEN_EPSILON
+    return float(total / (props.num_atoms + num_hydrogens))
+
+
+def _implicit_hydrogen_count(
```
pkacprzak5, line 508: I'm not sure about this function, I have come accross something like this:

The old eps_4 path failed when `build_reference_mol(..., saturated=True)` returned `None` (`SanitizeMol` / `AddHs`). The new path fails only when `_implicit_hydrogen_count` cannot fit `GetValenceList`.

Those are not the same gate. Example: an atom with an empty valence list (many metals) takes `_TAKES_HYDROGENS=False`, so this still returns a mean epsilon, while the old sanitizer would NaN the column. The other way around is also possible (table says no fit, RDKit would have sanitized).

Is the NaN pattern identical on the molecules where the old saturated reference failed? The 300-mol parity does not really cover that. A handful of metals / odd charges / hypervalent S,P against the old helper would be enough.


but I'm not quite sure whether that is the desired behavior, or should it be changed somehow?

tests/fingerprints/_new_mordred/extended_topochemical_atom.py
pkacprzak5, line 36: This helper does not follow `calculator.compute`.

Production: `props_regular` + `props_hydrogens` from `AddHs(mol_regular)` + kekulized bond types from `Kekulize(Mol(mol_regular), sanitize=False)` + `distance_matrix_regular` + `rings_regular`.

This test: everything from `preprocess_mol(mol, kekulize=True)` and a second H-explicit kekulized mol.

Could `_compute` copy the calculator wiring (or call `compute` and slice the ETA columns)? Otherwise a numbering / aromatic-flag / H-bond-direction mismatch only shows up in the full fingerprint.

pkacprzak5 (changes requested):

PR #673, opened 2026-08-09
Mordred: compute the autocorrelations from distance masks

j-adamczyk:
The autocorrelation descriptors looped over lags 0-8 in Python, and for each
lag over every atom pair, reading the atomic property of each atom through a
per-atom function. Every one of the eight weighting properties repeated that
whole traversal.

The properties now come from AtomicProperties as arrays, the distance matrix
is turned into one boolean mask per lag once, and each family (ATS, ATSC,
AATS, AATSC, MATS, GATS, Moreau, Geary) is a matrix product or a masked sum
over those.

Values are unchanged except for the centered families at the lags where the
result is zero by construction: 32 columns differ by at most 1e-13 absolute,
with both values below 1e-12 - floating point noise around zero, on 300
molecules sampled from MoleculeNet, TDC and MoleculeACE.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-08-eta` - review and merge in order.

<details>
<summary>The stack</summary>

2. #666 - AtomicProperties
5. #669 - RingSets
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations  **<- this PR**
10. #674 - Barysz matrices
13. #677 - walk counts
</details>

### DIFF
--- a/skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
+++ b/skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
@@ -1,12 +1,8 @@
 import numpy as np
-from rdkit.Chem import Mol
 
-from skfp.descriptors import atomic_partial_charges
 from skfp.fingerprints._new_mordred.utils.atomic_properties import (
-    PROPERTY_FUNCS,
-    get_intrinsic_state,
-    get_sigma_electrons,
-    get_valence_electrons,
+    WEIGHTING_PROPERTY_NAMES,
+    AtomicProperties,
 )
 from skfp.fingerprints._new_mordred.utils.graph_matrix import DistanceMatrix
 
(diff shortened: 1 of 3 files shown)


skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
```diff
@@ -59,149 +54,126 @@ def calc(mol_hydrogens: Mol, distance_matrix_hydrogens: DistanceMatrix) -> np.nd
     Following the original Mordred implementation, this function uses hydrogen-explicit
     molecule and distance matrix.
```
pkacprzak5, line 55: Maybe it is worth saying that this function does not use the molecule itself, and rather the `AtomicProperies` of it now?

skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
```diff
@@ -59,149 +54,126 @@ def calc(mol_hydrogens: Mol, distance_matrix_hydrogens: DistanceMatrix) -> np.nd
...
-    # allows bulk tensor computation
-    dist_stack = np.stack(
-        [(distance_matrix_hydrogens.matrix == d) for d in range(1, 9)], axis=0
+    dist_masks = np.stack(
+        [
+            (distance_matrix_hydrogens.matrix == dist)
+            for dist in range(1, MAX_DISTANCE + 1)
+        ],
+        axis=0,
+    ).astype(np.float64)
```
pkacprzak5, line 66: why do we use `np.float64` here? can we use maybe some other type that uses less memory? We used `bool` before, does the current later `einsum` need that or will it handle `bool`?

skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
```diff
@@ -59,149 +54,126 @@ def calc(mol_hydrogens: Mol, distance_matrix_hydrogens: DistanceMatrix) -> np.nd
...
-
-        ats_d = _weighted_sums(props, dist_stack)  # shape (8,)
-        aats_d = np.where(pair_counts != 0, ats_d / pair_counts, np.nan)
-
-        ats_values.extend(ats_d.tolist())
-        aats_values.extend(aats_d.tolist())
-
-    return ats_values, aats_values
+    # plain (uncentered) ATS and AATS do not use signed partial charge
+    is_charge = np.array(
```
pkacprzak5, line 81: I think it encodes the same thing that `_PROP_NAMES_NO_CHARGE` does, can we maybe  also compute it at module level?

Maybe we can have one mask and do it like this:
_IS_CHARGE = np.asarray(
    [name == "gasteiger_charge" for name in WEIGHTING_PROPERTY_NAMES]
)
_PROP_NAMES_NO_CHARGE = [
    name for name, is_c in zip(WEIGHTING_PROPERTY_NAMES, _IS_CHARGE) if not is_c
]

skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
```diff
@@ -59,149 +54,126 @@ def calc(mol_hydrogens: Mol, distance_matrix_hydrogens: DistanceMatrix) -> np.nd
...
+            0.5 * np.einsum("pdi,pi->pd", weighted_centered, props_centered),
+        ]
+    )
+    aatsc = _per_pair_average(atsc, pair_counts, num_atoms)
+
+    # MATS (Moran coefficient): the centered per-pair average, normalized by
+    # property variance around its mean
+    variation = centered_square_sums[:, np.newaxis]
+    mats = np.where(variation != 0, num_atoms * aatsc[:, 1:] / variation, np.nan)
+    mats = np.where(pair_counts != 0, mats, np.nan)
```
pkacprzak5, line 147: Isn't this `np.where` redundant, as the `_pair_pair_average` already does it? or do those things in between change it?

skfp/fingerprints/_new_mordred/descriptors/autocorrelation.py
```diff
@@ -59,149 +54,126 @@ def calc(mol_hydrogens: Mol, distance_matrix_hydrogens: DistanceMatrix) -> np.nd
...
-        if props_var != 0:
-            gats_d = mean_squared_diff / props_var
-        else:
-            gats_d = np.full_like(mean_squared_diff, np.nan)
-
-        gats_d = np.where(pair_counts != 0, gats_d, np.nan)
-
-        gats_values.extend(gats_d.tolist())
+    # M_d @ p for every property and distance, shape (n_props, 8, n)
+    weighted = np.einsum("dij,pj->pdi", dist_masks, props)
```
pkacprzak5, line 117: if this is just a simple matmul, wouldn't it be more readable with a simple `np.matmul(dist_masks, props.T)` and in the bottom accordingly? This made me wondering what exactly is happening there and why, with matmul i think it is easier to see that it is just simple M_d@p (i know there is a comment about it above, but i would still consider just standard matmul)

pkacprzak5 (changes requested):

PR #674, opened 2026-08-09
Mordred: build the eight Barysz matrices as one stack

j-adamczyk:
The Barysz descriptors build a weighted distance matrix per atomic property,
eight in all, and take the spectral attributes of each. Every matrix was built
by walking the bonds in Python, and each was decomposed on its own.

The weights of all eight are now computed together as one array indexed by
bond and property, the matrices are filled as a stack, and their attributes
come from a single spectral_attributes() call over that stack.

---

Splits up #661 into reviewable pieces. Based on `mordred-opt-09-autocorrelation` - review and merge in order.

<details>
<summary>The stack</summary>

2. #666 - AtomicProperties
5. #669 - RingSets
7. #671 - detour matrix
8. #672 - ETA descriptors
9. #673 - autocorrelations
10. #674 - Barysz matrices  **<- this PR**
13. #677 - walk counts
</details>

### DIFF
--- a/skfp/fingerprints/_new_mordred/descriptors/barysz_matrix.py
+++ b/skfp/fingerprints/_new_mordred/descriptors/barysz_matrix.py
@@ -1,13 +1,13 @@
 import numpy as np
-from rdkit.Chem import Mol
-from rdkit.Chem.rdchem import Atom
+from scipy.sparse import coo_matrix, csr_matrix
 from scipy.sparse.csgraph import floyd_warshall
 
 from skfp.fingerprints._new_mordred.utils.atomic_properties import (
-    PROPERTY_FUNCS,
+    CARBON_ELEMENT_PROPERTIES,
+    ELEMENT_PROPERTY_TABLES,
     AtomicProperties,
 )
-from skfp.fingerprints._new_mordred.utils.matrix_attributes import MatrixAttributes
+from skfp.fingerprints._new_mordred.utils.matrix_attributes import spectral_attributes
 
 """
 This code has been adapted from the BSD-licensed mordred-community library.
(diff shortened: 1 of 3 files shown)


skfp/fingerprints/_new_mordred/descriptors/barysz_matrix.py
```diff
@@ -48,76 +50,82 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
-        bo_arr = np.array([b.GetBondTypeAsDouble() for b in bonds])
-        weights = carbon_value**2 / (
-            property_values[i_arr] * property_values[j_arr] * bo_arr
+    diagonals = 1.0 - carbon_values.astype(np.float32) / prop_vals
+    return weights, diagonals
+
+
+class _BondGraph:
+    """
+    COO sparse matrix graph representation. Easy to add weighting by properties
```
pkacprzak5, line 109: This docstring looks weird, what do you mean by "easy to add weighting by properties"? I think this class should be described more precisely

skfp/fingerprints/_new_mordred/descriptors/barysz_matrix.py
```diff
@@ -48,76 +50,82 @@ def calc(mol_regular: Mol, props_regular: AtomicProperties, n_frags: int) -> np.
...
+    later.
+    """
+
+    def __init__(self, props: AtomicProperties):
+        self.shape = (props.num_atoms, props.num_atoms)
+        # compressing sorts the bonds by atom, which reorders their weights as well
+        pattern = coo_matrix(
```
pkacprzak5, line 116: i think that it should be an undirected graph? this way we make it directed, and only because we do floyd-warshall with directed=False it works, i think it is worth noting somewhere in comment

pkacprzak5 (changes requested):

PR #683, opened 2026-08-11
Add NN fingerprint base class

Thematiq:
## Changes

- Introduced `BaseNeuralFingerprintTransformer` as a base class for PyTorch based pre-trained fingerprint models. 
- Added GPU device acceleration to NN models.
- Updated CLAMP to utilize base class.

### DIFF
--- a/skfp/fingerprints/neural/base_neural_fp_transformer.py
+++ b/skfp/fingerprints/neural/base_neural_fp_transformer.py
@@ -1,0 +1,241 @@
+from __future__ import annotations
+
+from abc import abstractmethod
+from collections.abc import Sequence
+from copy import deepcopy
+
+import numpy as np
+import scipy.sparse
+import torch
+from huggingface_hub import hf_hub_download
+from joblib import effective_n_jobs
+from rdkit.Chem import Mol
+from scipy.sparse import csr_array
+from tqdm import tqdm
+
+from skfp.bases.base_fp_transformer import BaseFingerprintTransformer
+from skfp.utils import run_in_parallel
+
+_MODEL_CACHE: dict[tuple[type, str, str], object] = {}
+
+
+class BaseNeuralFingerprintTransformer(BaseFingerprintTransformer):
+    """
+    Base class for neural fingerprints.
+
+    Utilizes PyTorch-based pre-trained neural networks to compute fingerprints.
+
+    This class is not meant to be used directly. If you want to use custom neural fingerprints,
+    inherit from this class and implement the ``._prepare_input()`` and ``._forward_nn()`` methods.
+
+    Parameters
+    ----------
+    n_features_out : int
+        Number of output features.
+
+    requires_conformers : bool, default=False
+        Whether the fingerprint requires 3D conformations as inputs.
+
+    sparse : bool, default=False
+        Whether to return dense NumPy array, or sparse SciPy CSR array.
+
+    n_jobs : int, default=None
+        The number of jobs to run in parallel. :meth:`transform` is parallelized
+        over the input molecules when the device is CPU. ``None`` means 1 unless in a
+        :obj:`joblib.parallel_backend` context. ``-1`` means using all processors.
+        If the device is not CPU, the model runs all processing sequentially,
+        regardless of this value.
+        See scikit-learn documentation on ``n_jobs`` for more details.
+
+    batch_size : int, default=None
+        Number of inputs processed in each forward pass through the model.
+        ``None`` means all inputs are processed in a single pass when running
+        sequentially. When running with multiple CPU jobs, ``None``
+        instead divides the input into as many equal-sized parts as ``n_jobs``.
+
+    verbose : int or dict, default=0
+        Controls the verbosity when computing fingerprints.
+        If a dictionary is passed, it is treated as kwargs for ``tqdm()``,
+        and can be used to control the progress bar.
+
+    device : str or torch.device, default="cpu"
+        Device to use for the model inference. If the device is not CPU,
+        the model runs all processing sequentially, regardless of ``n_jobs``.
+        Use ``batch_size`` to bound memory usage on such devices.
+
+    weights_path : str or None, default=None
+        Path to a local pretrained checkpoint file (``.pt``).
+    """
+
+    _HF_REPO: str
+    _HF_FILENAME: str
+
+    # parameters common for all neural fingerprints
+    _parameter_constraints: dict = {
+        **BaseFingerprintTransformer._parameter_constraints,
+        "device": [str, torch.device],
+        "weights_path": [str, None],
+    }
+
+    def __init__(
+        self,
+        n_features_out: int,
+        requires_conformers: bool = False,
+        sparse: bool = False,
+        n_jobs: int | None = None,
+        batch_size: int | None = None,
+        verbose: int | dict = 0,
+        device: str | torch.device = "cpu",
+        weights_path: str | None = None,
+    ):
+        super().__init__(
+            n_features_out=n_features_out,
+            requires_conformers=requires_conformers,
+            sparse=sparse,
+            n_jobs=n_jobs,
+            batch_size=batch_size,
+            verbose=verbose,
+        )
+
+        self.device = device
+        self.weights_path = weights_path
+
+    def get_model(self):
+        """
+        Get the pre-trained model.
+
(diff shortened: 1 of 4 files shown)


skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,241 @@
+from __future__ import annotations
```
Thematiq: Added here instead of `skfp.bases` to keep all `scikit-fingeprints[neural]` modules in one place.

j-adamczyk: Add comment why this is necessary

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+from skfp.utils import run_in_parallel
+
+_MODEL_CACHE: dict[tuple[type, str, str], object] = {}
+
+
+class BaseNeuralFingerprintTransformer(BaseFingerprintTransformer):
+    """
+    Base class for neural fingerprints.
+
+    Utilizes PyTorch-based pre-trained neural networks to compute fingerprints.
```
j-adamczyk: Without "PyTorch", this is a technical detail. Maybe "Calculates embeddings from pretrained neural networks to be used as molecular fingeprints."

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+        If a dictionary is passed, it is treated as kwargs for ``tqdm()``,
+        and can be used to control the progress bar.
+
+    device : str or torch.device, default="cpu"
+        Device to use for the model inference. If the device is not CPU,
+        the model runs all processing sequentially, regardless of ``n_jobs``.
+        Use ``batch_size`` to bound memory usage on such devices.
+
+    weights_path : str or None, default=None
+        Path to a local pretrained checkpoint file (``.pt``).
```
j-adamczyk: What happens is this is `None`? Describe in docstring. Also, .pt in parentheses looks a bit ugly, and is not strictly necessary, e.g. for HuggingFace models which can have many shards

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+        sequentially. When running with multiple CPU jobs, ``None``
+        instead divides the input into as many equal-sized parts as ``n_jobs``.
+
+    verbose : int or dict, default=0
+        Controls the verbosity when computing fingerprints.
+        If a dictionary is passed, it is treated as kwargs for ``tqdm()``,
+        and can be used to control the progress bar.
+
+    device : str or torch.device, default="cpu"
+        Device to use for the model inference. If the device is not CPU,
```
j-adamczyk: Provide an example of alternative, e.g. "cuda" or "cuda:0"

skfp/fingerprints/neural/base_neural_fp_transformer.py
j-adamczyk: Maybe also mention that we assume that models are hosted on HuggingFace? It would be quite challenging to inherit from this and use non-HF model

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+        Get the pre-trained model.
+
+        The return type depends on the subclass — it does not have to be a
+        ``torch.nn.Module``, only something ``_forward_nn()`` knows how to call.
+
+        Returns
+        -------
+        model
+            The pretrained model, as constructed by ``_load_model()``, moved
+            to ``self.device``, and cached.
```
j-adamczyk: I'm not sure about mentioning private methods in public method docstrings. If those methods are relevant, maybe something private should be public, or vice versa?

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        path = self.weights_path or hf_hub_download(
+            repo_id=self._HF_REPO, filename=self._HF_FILENAME
+        )
+        return self._get_cached_model(path, self.device)
+
+    @classmethod
+    def _get_cached_model(cls, path: str, device: str | torch.device):
```
j-adamczyk: Private methods generally should be at the bottom of the class. Maybe we can make an exception just for abstract methods, since they are important to override when inheriting.

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+    def _to_device(self, X):
+        if isinstance(X, np.ndarray):
+            return torch.from_numpy(X).to(self.device)
+        if hasattr(X, "to"):
+            return X.to(self.device)
+        if isinstance(X, dict) and all(hasattr(x, "to") for x in X.values()):
+            return {k: v.to(self.device) for k, v in X.items()}
+        raise RuntimeError("Cannot convert NN model to requested device")
+
+    def _calculate_fingerprint(self, X: Sequence[str | Mol]) -> np.ndarray | csr_array:
```
j-adamczyk: This method is very relevant to understand the methods used inside and their order. Put this above those methods. Also, you can then move much of their docstrings here, so user immediately sees what needs to be inherited and implemented, and how that is used.

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,239 @@
...
+        with torch.no_grad():
+            X_input = self._prepare_input(X)
+            X_input = self._to_device(X_input)
+            X_output = self._forward_nn(X_input)
+        if isinstance(X_output, torch.Tensor):
+            X_output = X_output.detach().cpu().numpy()
```
j-adamczyk: This should be in the docstring somewhere. Enforcing CPU sync + NumPy conversion is reasonable, but we should make it explicit.

skfp/fingerprints/neural/clamp.py
```diff
@@ -1,14 +1,16 @@
+from __future__ import annotations
```
j-adamczyk: Does this need to be added to every neural fingerprint module?

j-adamczyk (changes requested):

skfp/fingerprints/neural/clamp.py
Thematiq: Checked Sphinx. If we were to use `str | torch.device` in the doctorings then yes, and there's no way to omit it. But I've found an alternative approach with `Union[str, "torch.device"]` which we can use to annotate instead. `str | "torch.device"` won't work either as it is evaluated eagerly. But the `Union` renders as desired in docs:

Image

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,268 @@
+"""
+Base class for neural fingerprints. Stored in ``skfp.fingerprints.neural``
+instead of ``skfp.bases`` to avoid optional dependencies outside
+of the ``skfp.fingerprints.neural`` module.
+"""
```
j-adamczyk: We don't use module docstrings

skfp/fingerprints/neural/base_neural_fp_transformer.py
```diff
@@ -0,0 +1,268 @@
...
+# docs/conf.py), which replaces `torch.device`, `torch.Tensor`, etc. with
+# mock *instances* rather than classes. Python's `|` operator is evaluated
+# eagerly at class-definition time and has no special handling for mock
+# objects, so combining one via `|` (e.g. `str | torch.device`) raises a
+# `TypeError` under the mock. A bare, non-union annotation is fine either way.
+#
+# Fix by deferring evaluation: either `from __future__ import annotations`
+# at the top of the file, or `typing.Union[...]` with the torch side quoted
+# as a forward reference, as done below. Quoting only PART of a live `|`
+# expression (e.g. `str | "torch.device"`) does NOT work — the operator has
```
j-adamczyk: Do not use em dashes

j-adamczyk (changes requested):

j-adamczyk (approved):

PR #685, opened 2026-08-30
Mordred Lipinski

FloudMe77:
## Changes

Add Mordred Lipinski descriptors

Adding `Lipinski` / `GhoseFilter` needed LogP, molar refractivity, exact MW, atom
count and the H-bond donor/acceptor counts — all of which `calc_rdkit_2d` already
computed for `SLogP`, `SMR`, `MW`, `AMW`, `nHBAcc` and `nHBDon`.

Rather than recompute them (`Crippen.MolLogP`/`MolMR` and `CalcNumHBA`/`CalcNumHBD`
are SMARTS-matching passes over the whole molecule), they moved into
`MolecularProperties`, a shared dependency built once per molecule in
`calculator.compute()` and passed to both modules — the same pattern as
`AtomicProperties`, `DistanceMatrix` and `RingSets`.

So `calc_rdkit_2d` takes a third argument and reads those six values off it instead
of calling RDKit itself; `_average_exact_mol_wt` reduces to a division, still wrapped
in `safe_value` for the empty-molecule case. No descriptor values change — verified
against mordred-community.


skfp/fingerprints/_new_mordred/descriptors/mol_filters.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+Drug-likeness rule descriptors: the Lipinski rule of five and the Ghose filter.
+
+This code has been adapted from the BSD-licensed mordred-community library.
+https://github.com/JacksonBurns/mordred-community
+
+See skfp/fingerprints/data/mordred-community_bsd_license.txt for the license text.
+"""
+
+FEATURE_NAMES = ["Lipinski", "GhoseFilter"]
```
j-adamczyk, line 16: I would call this module "mol_filters" or similar

skfp/fingerprints/_new_mordred/descriptors/mol_filters.py
```diff
@@ -0,0 +1,43 @@
...
+    acceptors, molecular weight up to 500 and Wildman-Crippen LogP up to 5.
+
+    ``GhoseFilter`` bounds molecular weight to [160, 480], atom count (hydrogens
+    included) to [20, 70], LogP to [-0.4, 5.6] and molar refractivity to [40, 130].
+    """
+    lipinski_rule_of_five = (
+        mol_properties.num_h_bond_donors <= 5
+        and mol_properties.num_h_bond_acceptors <= 10
+        and mol_properties.exact_mol_wt <= 500
+        and mol_properties.log_p <= 5
```
j-adamczyk, line 34: Look at Lipinski filter in skfp, it uses slightly different functions for HBA and HBD

skfp/fingerprints/_new_mordred/utils/molecular_properties.py
```diff
@@ -0,0 +1,53 @@
...
+        self.log_p = log_p
+        self.molar_refractivity = molar_refractivity
+        self.exact_mol_wt = exact_mol_wt
+        # hydrogens included, whether they are explicit atoms or not
+        self.num_atoms = num_atoms
+        self.num_h_bond_acceptors = num_h_bond_acceptors
+        self.num_h_bond_donors = num_h_bond_donors
+
+    @classmethod
+    def from_mol(cls, mol: Mol) -> "MolecularProperties":
```
j-adamczyk: Instead, you can pass Mol to constructor and calculate this in init

j-adamczyk (changes requested):

skfp/fingerprints/_new_mordred/descriptors/mol_filters.py
FloudMe77: The original Mordred Lipinski descriptor uses CalcNumHBA / CalcNumHBD for HBA/HBD (through its HBondAcceptor/HBondDonor wrappers), so that's what I used here to match it. skfp's LipinskiFilter uses CalcNumLipinskiHBA / CalcNumLipinskiHBD instead, which are different functions.

I checked how much they actually differ on 2039 BBBP molecules: HBA differs on 52.5% of them, HBD on 18.7%, and the final Lipinski value flips on 1.72%. So the two really aren't interchangeable here

j-adamczyk (approved):

PR #688, opened 2026-08-31
Mordred: add the molecular ID descriptors

FloudMe77:
## Changes

Ports mordred's `MolecularId`: `MID` and `AMID` over all atoms, heteroatoms,
carbons, nitrogens, oxygens and halogens. The feature names were already in
`feature_names.py` and returned NaN until now.

Three commits: the descriptor, a follow-up trimming its fixed per-molecule
cost, and a docstring explaining the algorithmic cost for future readers.

## Why this one is expensive

An atomic ID sums the inverse square root of the edge weight product over every
simple path starting at that atom. The search is a backtracking walk, not a
plain DFS — the visited mark means "on the current path" and is undone on the
way out — so it runs once per simple path rather than once per atom.

The cost is therefore exponential in the number of **rings**, not in molecule
size. For an acyclic molecule there is exactly one path per ordered pair of
atoms, so the walk degenerates to a plain traversal; each independent cycle
multiplies the number of routes:

| molecule | atoms | rings | calls from one atom |
|---|---|---|---|
| octane | 8 | 0 | 8 (same as a plain DFS) |
| benzene | 6 | 1 | 11 |
| naphthalene | 10 | 2 | 32 |
| pentacene | 22 | 5 | 359 |
| cubane cage | 9 | 5 | 151 |

There is no value-preserving shortcut: the sum needs one term per path, and the
paths must be simple, so the state is the whole visited set rather than the
current atom — no per-atom dynamic programming can stand in for walking them.
This makes it the most expensive descriptor in the package.

## Performance

The arithmetic matches mordred exactly. The speedup comes from constants: the
graph is nested lists of plain ints instead of a networkx graph (which cost two
dict lookups per edge), visited atoms live in a `bytearray` instead of a `set`,
and partial sums are returned rather than accumulated on an object.

Best of N runs after warm-up:

| molecule | atoms | mordred | this | |
|---|---|---|---|---|
| Digoxin | 55 | 120.2 ms | 26.6 ms | 4.6× |
| pentacene | 22 | 12.8 ms | 2.5 ms | 5.4× |
| EllagicAcid | 22 | 9.6 ms | 2.0 ms | 5.7× |
| Lycopene | 40 | 3.6 ms | 0.59 ms | 6.2× |
| Caffeine | 14 | 1.6 ms | 0.26 ms | 6.2× |
| Benzene | 6 | 0.41 ms | 0.05 ms | 8.8× |

The second commit targets small molecules, where a size-independent ~35 µs of
mask building dominated everything else. Selecting halogens with `np.isin` alone
took 17.9 µs on a 12-atom molecule — sorting its handful of needles costs more
than the lookup — against 0.33 µs for a boolean table indexed by atomic number.
That, plus dropping the all-true mask for the first feature pair, takes the
fixed part from ~35 µs to ~15 µs: ethane goes from 43.6 µs to 24.3 µs, benzene
from 60.8 µs to 46.6 µs. Larger molecules are unaffected, being dominated by
the path search.

## Correctness

Verified against mordred on 35 molecules (the bundled reference set plus an
aminoglycoside, a fused polycyclic and inositol): worst relative difference
**5.77e-08**, below float32 resolution, and unchanged by the optimisation
commit. Disconnected molecules give NaN on both sides, matching mordred's
`require_connected`.

## Notes for reviewers

A heteroatom here is anything but carbon **and hydrogen**, unlike
`AtomicProperties.is_hetero`, which is carbon-only. This differs only for the
hydrogens `RemoveHs` keeps — `[H][H]` has `MID_h == 0`, where the shared
predicate would give 3.0 — and two tests pin it.

Reference values for `MID`/`AMID` are converted from mordred-community's own
`tests/references/MolecularId.yaml`. The element-filtered features are not
covered there, so those values were computed with mordred-community; the test
module's docstring records both provenances.

### DIFF
--- a/skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
+++ b/skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
@@ -38,7 +38,13 @@ _HYDROGEN_ATOMIC_NUM = 1
 _CARBON_ATOMIC_NUM = 6
 _NITROGEN_ATOMIC_NUM = 7
 _OXYGEN_ATOMIC_NUM = 8
-_HALOGEN_ATOMIC_NUMS = np.asarray(sorted(HALOGEN_ATOMIC_NUMS), dtype=np.intp)
+
+# one flag per atomic number, indexed by the atomic number itself. Selecting the
+# halogens this way is a plain gather, where np.isin spends more time sorting its
+# few needles than the whole lookup takes.
+_NUM_ATOMIC_NUMBERS = 119  # atomic numbers 0 to 118, as RDKit reports them
+_IS_HALOGEN = np.zeros(_NUM_ATOMIC_NUMBERS, dtype=bool)
+_IS_HALOGEN[sorted(HALOGEN_ATOMIC_NUMS)] = True
 
 
 def calc(props_regular: AtomicProperties, n_frags: int) -> np.ndarray:
@@ -62,18 +68,17 @@ def calc(props_regular: AtomicProperties, n_frags: int) -> np.ndarray:
     is_hetero = (atomic_nums != _HYDROGEN_ATOMIC_NUM) & (
         atomic_nums != _CARBON_ATOMIC_NUM
     )
-    selections = [
-        np.ones(props_regular.num_atoms, dtype=bool),
-        is_hetero,
-        atomic_nums == _CARBON_ATOMIC_NUM,
-        atomic_nums == _NITROGEN_ATOMIC_NUM,
-        atomic_nums == _OXYGEN_ATOMIC_NUM,
-        np.isin(atomic_nums, _HALOGEN_ATOMIC_NUMS),
+    molecular_ids = [
+        atom_ids.sum(),  # every atom, so no selection to build
+        atom_ids[is_hetero].sum(),
+        atom_ids[atomic_nums == _CARBON_ATOMIC_NUM].sum(),
+        atom_ids[atomic_nums == _NITROGEN_ATOMIC_NUM].sum(),
+        atom_ids[atomic_nums == _OXYGEN_ATOMIC_NUM].sum(),
+        atom_ids[_IS_HALOGEN[atomic_nums]].sum(),
     ]
 
     values = []
-    for selection in selections:
-        molecular_id = atom_ids[selection].sum()
+    for molecular_id in molecular_ids:
         # averaged over every atom of the molecule, not over the selected ones
         values += [molecular_id, molecular_id / props_regular.num_atoms]
 
--- a/skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
+++ b/skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
@@ -90,6 +90,9 @@ def _atom_ids(props: AtomicProperties) -> np.ndarray:
     Atomic ID of every atom: one plus half the sum, over all simple paths
     starting at that atom, of the inverse square root of the product of the
     path's edge weights.
+
+    This is by far the most expensive part of the descriptor; see
+    :func:`_sum_over_paths` for why enumerating those paths is exponential.
     """
     adjacency = _weighted_adjacency(props)
     num_atoms = props.num_atoms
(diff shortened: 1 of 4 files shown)


skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
```diff
@@ -0,0 +1,170 @@
...
+
+# a path stops being extended once the product of its edge weights reaches this
+# limit, as anything longer would contribute less than Mordred's 1e-10 epsilon;
+# Mordred spells the same number as int(1 / eps ** 2)
+_WEIGHT_PRODUCT_LIMIT = 10**20
+
+_HYDROGEN_ATOMIC_NUM = 1
+_CARBON_ATOMIC_NUM = 6
+_NITROGEN_ATOMIC_NUM = 7
+_OXYGEN_ATOMIC_NUM = 8
```
j-adamczyk: I would use those directly, they are common knowledge

skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
```diff
@@ -0,0 +1,170 @@
...
+
+# one flag per atomic number, indexed by the atomic number itself. Selecting the
+# halogens this way is a plain gather, where np.isin spends more time sorting its
+# few needles than the whole lookup takes.
```
j-adamczyk: Shorter: "fast lookup, vectorized in NumPy"

skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def _atom_ids(props: AtomicProperties) -> np.ndarray:
+    """
+    Atomic ID of every atom: one plus half the sum, over all simple paths
+    starting at that atom, of the inverse square root of the product of the
+    path's edge weights.
+
+    This is by far the most expensive part of the descriptor; see
+    :func:`_sum_over_paths` for why enumerating those paths is exponential.
```
j-adamczyk: Unnecessary

skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
```diff
@@ -0,0 +1,170 @@
...
+    degrees of the two bonded atoms.
+
+    Kept as nested Python lists of plain ints rather than NumPy arrays: the path
+    search below reads them one edge at a time, where boxing NumPy scalars costs
+    far more than the lookup itself.
+    """
+    degrees = props.degrees
+    begin_idxs = props.bond_begin_idxs.tolist()
+    end_idxs = props.bond_end_idxs.tolist()
+    weights = (degrees[props.bond_begin_idxs] * degrees[props.bond_end_idxs]).tolist()
```
j-adamczyk, line 107: Why .tolist()? I think you can just iterate here

skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
```diff
@@ -0,0 +1,170 @@
...
+    This is a backtracking search rather than a plain depth-first traversal:
+    ``visited`` marks the atoms on the *current path*, not the atoms already
+    seen, and the mark is undone on the way out. An atom is therefore re-entered
+    once per route that reaches it, and the recursion has one call per simple
+    path instead of one per atom. The cost is exponential in the number of rings
+    -- a plain traversal of pentacene makes 22 calls from a given atom, this one
+    makes 359 -- and there is no way around it: the sum needs one term per path,
+    and the paths have to be simple, so no per-atom bookkeeping can stand in for
+    walking them. It does degenerate to a plain traversal for acyclic molecules,
+    where only one route reaches each atom.
```
j-adamczyk: Simplify, make less LLM-sounding

j-adamczyk (changes requested):

skfp/fingerprints/_new_mordred/descriptors/molecular_id.py
FloudMe77: The products overflow with NumPy scalars. _sum_over_paths multiplies these edge weights along each path up to _WEIGHT_PRODUCT_LIMIT = 1e20, which is past int64's max (~9.2e18). Original Mordred does this with arbitrary-precision Python ints (its cutoff is the same int(1/eps²) = 1e20), and .tolist() is what keeps that behaviour here. If I iterate the arrays directly, weight_product * weight overflows to a negative number and math.sqrt then throws math domain error — it crashes on larger molecules like a C40 chain or a big fused ring system. So this isn't just a speed thing, it's needed for correctness. I'll reword the docstring to say that, since right now it only mentions the boxing cost.

PR #689, opened 2026-08-31
Mordred: add the molecular framework ratio descriptor

FloudMe77:
## Changes

Ports mordred's `Framework`: `fMF`, the share of the molecule made up of its ring
atoms plus the linkers joining them. The feature name was already in
`feature_names.py` and returned NaN until now.

Every input is already computed and shared, so the descriptor adds no new
per-molecule work of its own: the SSSR rings come from `RingSets`, the bonds from
`AtomicProperties`, and the denominator is `props_hydrogens.num_atoms`.

## Approach

Each ring is contracted to a single node, so a route between two rings is a path
in that quotient graph, and the non-ring atoms on it are the linkers. Three
changes make this **18–43× faster** than mordred:

- **one search per ring instead of one per pair.** A breadth-first search from a
  ring reaches every other ring at once; mordred makes a separate
  `nx.shortest_path` call for each of the `NR·(NR−1)/2` pairs. Digoxin: 8
  searches instead of 28.
- **no networkx** — flat adjacency lists and a `deque`.
- **hydrogens stay out of the graph.** Mordred iterates the bonds of the
  hydrogen-explicit molecule, which adds dozens of leaves. A hydrogen has one
  bond, so it can only ever be a dead end, never an interior atom of a path
  between two rings.

Median of repeated runs after warm-up:

| molecule | rings | mordred | this | |
|---|---|---|---|---|
| Digoxin | 8 | 1331.7 µs | 71.9 µs | 18.5× |
| Epicatechin | 3 | 343.4 µs | 14.8 µs | 23.2× |
| Caffeine | 2 | 238.8 µs | 9.9 µs | 24.1× |
| fused pentacyclic | 5 | 852.8 µs | 19.9 µs | 42.9× |
| two rings + linker | 2 | 445.7 µs | 10.7 µs | 41.6× |

## Why picking one shortest path is safe

Mordred takes a single shortest path per pair, so in principle the linker set
could depend on which one networkx happens to return. It cannot: a second,
different route between two rings would close a cycle, and ring perception turns
that cycle into a ring, so its atoms stop being linkers in the first place. I
checked every ring pair in the reference set and in four deliberately bridged
structures ([2.2]paracyclophane and friends) — zero ambiguous pairs, and in the
bridged cases every atom ends up a ring atom, leaving nothing to disambiguate.
The reasoning is recorded in `_linker_atoms`'s docstring.

## Correctness

Verified against mordred end-to-end through `NewMordredFingerprint` on 37
molecules: worst relative difference **5.02e-08**, below float32 resolution.

Reference values come from mordred-community's own
`tests/references/Framework.yaml`; Cyanidin is commented out there and is left
out here too. Thirteen further cases were computed with mordred-community and
cover the parts the reference set does not exercise: no rings at all, a lone ring
with a dangling chain (**not** a linker — a linker needs two rings), fused and
directly bonded rings (no linkers), one- and four-atom linkers, a disconnected
molecule whose two rings have no path between them, and the empty molecule.

### DIFF
--- a/skfp/fingerprints/_new_mordred/descriptors/framework.py
+++ b/skfp/fingerprints/_new_mordred/descriptors/framework.py
@@ -1,0 +1,124 @@
+from collections import deque
+
+import numpy as np
+
+from skfp.fingerprints._new_mordred.descriptors.ring_count import RingSets
+from skfp.fingerprints._new_mordred.utils.atomic_properties import AtomicProperties
+
+"""
+Molecular framework ratio descriptor.
+
+References
+    * https://doi.org/10.1021/jm9602928
+
+This code has been adapted from the BSD-licensed mordred-community library.
+https://github.com/JacksonBurns/mordred-community
+
+See skfp/fingerprints/data/mordred-community_bsd_license.txt for the license text.
+"""
+
+FEATURE_NAMES = ["fMF"]
+
+
+def calc(
+    props_regular: AtomicProperties,
+    rings_regular: RingSets,
+    num_atoms_hydrogens: int,
+) -> np.ndarray:
+    """
+    Compute the molecular framework ratio, the share of the molecule taken up by
+    its framework: the ring atoms together with the linkers joining them.
+
+    The denominator counts hydrogens, so benzene scores 6 / 12 rather than 1.
+    """
+    if num_atoms_hydrogens == 0:
+        return np.full(len(FEATURE_NAMES), np.nan, dtype=np.float32)
+
+    ring_atom_sets = rings_regular.simple_ring_atom_sets
+    framework_atoms = {atom for ring in ring_atom_sets for atom in ring}
+    framework_atoms |= _linker_atoms(props_regular, ring_atom_sets)
+
+    return np.asarray([len(framework_atoms) / num_atoms_hydrogens], dtype=np.float32)
+
+
+def _linker_atoms(props: AtomicProperties, ring_atom_sets: list[set[int]]) -> set[int]:
+    """
+    Atoms outside every ring that lie on a shortest path between two rings.
+
+    Each ring is contracted to a single node, so a route between two rings is a
+    path in that quotient graph. Mordred searches once per pair of rings; a
+    breadth-first search from one ring reaches all the others at once, so one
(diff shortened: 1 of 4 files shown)


skfp/fingerprints/_new_mordred/descriptors/framework.py
```diff
@@ -0,0 +1,124 @@
...
+import numpy as np
+
+from skfp.fingerprints._new_mordred.descriptors.ring_count import RingSets
+from skfp.fingerprints._new_mordred.utils.atomic_properties import AtomicProperties
+
+"""
+Molecular framework ratio descriptor.
+
+References
+    * https://doi.org/10.1021/jm9602928
```
j-adamczyk: Move to calc() description

skfp/fingerprints/_new_mordred/descriptors/framework.py
```diff
@@ -0,0 +1,124 @@
...
+    The denominator counts hydrogens, so benzene scores 6 / 12 rather than 1.
+    """
+    if num_atoms_hydrogens == 0:
+        return np.full(len(FEATURE_NAMES), np.nan, dtype=np.float32)
+
+    ring_atom_sets = rings_regular.simple_ring_atom_sets
+    framework_atoms = {atom for ring in ring_atom_sets for atom in ring}
+    framework_atoms |= _linker_atoms(props_regular, ring_atom_sets)
+
+    return np.asarray([len(framework_atoms) / num_atoms_hydrogens], dtype=np.float32)
```
j-adamczyk, line 42: Calculate this value above as a variable

skfp/fingerprints/_new_mordred/descriptors/framework.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _linker_atoms(props: AtomicProperties, ring_atom_sets: list[set[int]]) -> set[int]:
+    """
+    Atoms outside every ring that lie on a shortest path between two rings.
+
+    Each ring is contracted to a single node, so a route between two rings is a
```
j-adamczyk, line 49: Too much detail on Mordred implementation - we care about what is here

j-adamczyk (changes requested):

